PR #4, opened 2017-04-20
Add support for building openSUSE Leap 42.2 stemcells

aduffeck:
This pull requests adds support for building an openSUSE Leap 42.2 stemcell. Additionally, the pull request includes some changes which are required for compatibility with openSUSE and generic improvements which make the code more robust.

Our changes integrate closely with the upstream structure. Building the openSUSE stemcell requires a custom Docker container because it relies on Kiwi[1] for creating the root file system, which is the standard way of building SUSE base images.

This pull request depends on the pull request[2] which makes the bosh-agent compatible with openSUSE.

[1] https://doc.opensuse.org/projects/kiwi/doc/
[2] https://github.com/cloudfoundry/bosh-agent/pull/127


bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
```diff
@@ -286,6 +288,28 @@ def photonos_os_stages
...
+        bosh_steps,
+        :password_policies,
+        :restrict_su_command,
+        :tty_config,
+        :rsyslog_config,
+        :delay_monit_start,
+        :system_grub,
+        :cron_config,
+        :system_users,
+        :bosh_audit_opensuse,
```
cppforlife: any reason why not to include `escape_ctrl_alt_del` stage?

aduffeck: No, that just looks like an omission on our side. I pushed a commit which includes that stage.

cppforlife:
story: https://www.pivotaltracker.com/story/show/146464235

stemcell_builder/stages/untar_base_os_image/apply.sh
```diff
@@ -6,7 +6,7 @@ base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
 
 # the tar_base_os_image stage creates a tar ball of the chroot with no enclosing directory
-tar zxf $os_image_tgz -C $chroot
+sudo tar zxf $os_image_tgz -C $chroot --same-owner
```
tjvman: Per the [tar man page](https://linux.die.net/man/1/tar), `--same-owner` is the default behavior for the superuser, so don't need to specify that flag.

bosh-stemcell/spec/support/stemcell_shared_examples.rb
```diff
@@ -37,8 +37,8 @@
       its (:stdout) { should eq('') }
     end
 
-    describe file('/etc/hosts.equiv') do
-      it { should_not be_file }
+    describe command('[ ! -e /etc/hosts.equiv ] || egrep \^[^#].+ /etc/hosts.equiv') do
```
tjvman: [STIG V38491](https://www.stigviewer.com/stig/red_hat_enterprise_linux_6/2013-06-03/finding/V-38491) mandates that `/etc/hosts.equiv` must not exist at all on the system, so although an empty file would probably not be an issue in practice, it would be in violation of that STIG's requirements. 

Should switch back to the previous behavior - ruby's [File.file?](https://ruby-doc.org/core-2.2.0/File.html#method-c-file-3F) will ensure that the file is not present, and we prefer using ruby libraries rather than shell commands where possible.

stemcell_builder/stages/base_opensuse/apply.sh
```diff
@@ -0,0 +1,95 @@
...
+# This is required for the bosh_go_agent stage
+mkdir -p $chroot/etc/service/
+
+run_in_chroot $chroot "
+  # Install runit from sources
+  mkdir -p /package
+  chmod 1755 /package
+  cd /package
+
+  wget http://smarden.org/runit/runit-2.1.2.tar.gz
```
tjvman: I'd feel more comfortable if we were downloading runit from somewhere that provides either shasums or sig files so that we can verify this package.

bosh-stemcell/spec/os_image/opensuse_leap_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+  end
+
+  context 'official Centos gpg key is installed (stig: V-38476)' do
+    describe command('rpm -qa gpg-pubkey* 2>/dev/null | xargs rpm -qi 2>/dev/null') do
+      its(:stdout) { should include('SuSE Package Signing Key') }
+    end
+  end
+
+  context 'gpgcheck must be enabled (stig: V-38483)' do
+    describe file('/etc/zypp/zypp.conf') do
```
tjvman: would prefer to have this be a positive test (check that gpgcheck, etc are set to a positive boolean value)

stemcell_builder/stages/base_opensuse/apply.sh
aduffeck: Good point, unfortunately I couldn't find any somewhat-official place where this would be the case.

Would you be fine with shipping the sources as an asset instead (~109K)?

tjvman: Would it be possible to install this specific runit version from an official OpenSUSE repository? For example, we use `apt` to install runit for Ubuntu stemcells.

ci/docker/suse-os-image-stemcell-builder/scripts/install-go.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -eux
+
+GOPATH=/home/opensuse/go
+GO_ARCHIVE_URL=https://storage.googleapis.com/golang/go1.6.1.linux-amd64.tar.gz
```
tjvman: Is there a reason for using this specific version of golang?

aduffeck: I don't think so, I bumped it to 1.8.3 and it's still working fine.

stemcell_builder/stages/base_ssh/apply.sh
```diff
@@ -80,6 +83,10 @@ elif [ "$(get_os_type)" == "ubuntu" ]; then
   sed "/^ *MACs/d" -i $chroot/etc/ssh/sshd_config
   echo 'MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-ripemd160-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,hmac-ripemd160,hmac-sha1' >> $chroot/etc/ssh/sshd_config
 
+elif [ "$(get_os_type)" == "opensuse" ]; then
```
tjvman: This `elif` will never get evaluated because the `if` on line 68 will evaluate to true when using opensuse.

bosh-stemcell/spec/support/stemcell_opensuse_shared_examples.rb
```diff
@@ -0,0 +1,134 @@
+shared_examples_for 'a openSUSE stemcell' do
+
+  describe command('ls -1 /lib/modules | wc -l') do
+    its(:stdout) {should eq "1\n"}
+  end
+
+  context 'installed by image_install_grub', exclude_on_warden: true do
```
tjvman: This exclusion breaks the tests in `cis_spec.rb` and `stig_spec.rb`. Removing it will cause those tests to pass.

bosh-stemcell/spec/support/stemcell_opensuse_shared_examples.rb
```diff
@@ -0,0 +1,134 @@
...
+    exclude_on_warden: true,
+    exclude_on_azure: true,
+   } do
+    describe file('/var/vcap/bosh/agent.json') do
+      it { should be_valid_json_file }
+      its(:content) { should match('"Type": "CDROM"') }
+    end
+  end
+
+  context 'installed by bosh_azure_agent_settings', {
```
tjvman: Should have tests for `bosh_softlayer_agent_settings` and `bosh_openstack_agent_settings` as well.

tjvman:
Thanks for submitting this PR! Excited to see a new OS available for BOSH stemcells. There's some things we'd like to see fixed before merging this in:
* When a `create-env` director VM's agent is bootstrapping, it is failing to bring up the `eth0` interface, which causes the deploy to fail because networking is unavailable. May also be true for non-`create-env` machines.
* The version of `sfdisk` in the openSUSE image is incompatible with the BOSH agent. The most straightforward fix would be to explicitly set `parted` as the partitioner (as was done for AWS and Openstack in the changeset).
* The `suse-os-image-stemcell-builder` doesn't install OVFtools, which forces users to switch docker images in order to build vSphere stemcells. 
* In order to comply with security benchmarking, only `vcap` and `root` should have login-able shells. All other users should have password login disabled and non login-able shells. All groups should have password login disabled. There should be an explicit test of these behaviors, as well.

Have these changes been tested on all the BOSH-supported infrastructures? Should be able to deploy a BOSH director with `create-env`, deploy [zookeeper](https://github.com/cppforlife/zookeeper-release) and run the `tests/run.sh` script successfully for each IaaS.
Just for the sake of reference, here's the list: 
* AWS
* Azure 
* GCP
* OpenStack
* vSphere
* Warden - The host (i.e., the `create-env` director) can be running on Ubuntu, the containers should be running openSUSE. Support for Warden allow users to test against openSUSE much more easily than deploying full VMs.

[bosh-deployment](github.com/cloudfoundry/bosh-deployment) has manifests and ops-files for all the above IaaSes, but note that it's using releases precompiled on Ubuntu stemcells, which probably won't work with openSUSE. You should be able to use the `local-bosh-release.yml` ops-file to deploy a non-compiled release of BOSH; for the others you'll need to change them manually.

greygoo:
Hi, sorry for the short delay, I'm also working on the openSUSE stemcells and just wanted to let you know we are working on the suggested changes and will update the PR in one of our next sprints.
About testing the supported infrastructures - we are focused on OpenStack testing at the moment but will run tests on other environments suggested as well.

bosh-stemcell/spec/os_image/opensuse_leap_spec.rb
ericpromislow: This was done in 2878d9151690a90052683c9c7fba78e41098acaf

ericpromislow:
Here's an update on the PR, first addressing the comments from https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/4#issuecomment-310215238

    Thanks for submitting this PR! Excited to see a new OS available for BOSH stemcells. There's some things we'd like to see fixed before merging this in:

    * When a create-env director VM's agent is bootstrapping, it is failing to bring up the eth0 interface, which causes the deploy to fail because networking is unavailable. May also be true for non-create-env machines.

No longer observed

    * The version of sfdisk in the openSUSE image is incompatible with the BOSH agent. The most straightforward fix would be to explicitly set parted as the partitioner (as was done for AWS and Openstack in the changeset).

Done

    * The suse-os-image-stemcell-builder doesn't install OVFtools, which forces users to switch docker images in order to build vSphere stemcells.

Done in the opensuse container

    * In order to comply with security benchmarking, only vcap and root should have login-able shells. All other users should have password login disabled and non login-able shells. All groups should have password login disabled. There should be an explicit test of these behaviors, as well.

Done

We still haven't tested on that full list of IaaSes.

monkeyherder:
cc @tjvman @dpb587-pivotal Any other input on this before we pull it in?

tjvman:
@monkeyherder LGTM

stemcell_builder/stages/bosh_go_agent/apply.sh
```diff
@@ -26,11 +26,11 @@ cp -a $assets_dir/alerts.monitrc $chroot/var/vcap/monit/alerts.monitrc
...
 if is_ppc64le; then
-  curl -L -o bosh-agent "https://s3.amazonaws.com/bosh-agent-binaries/bosh-agent-2.15.0-linux-ppc64le?versionId=GS9O3Z13bjjPYWQXyKnt1EH5OuJvQrP4"
-  echo "d034db7caa859ad64ea35bd262ac87ea9c24c21aed1f30c4f749b49bfa11aa96  bosh-agent" | shasum -a 256 -c -
+  curl -L -o bosh-agent "https://s3.amazonaws.com/bosh-agent-binaries/bosh-agent-2.29.0-linux-ppc64le?versionId=cyyDKlnKwwW227tJ.W0eI9PtSqs0dotS"
+  echo "23464436e8249e2d5dae66d7d0b6ea533d9d06ef889263cd000502fb52e70764  bosh-agent" | shasum -a 256 -c -
 else
-  curl -L -o bosh-agent "https://s3.amazonaws.com/bosh-agent-binaries/bosh-agent-2.15.0-linux-amd64?versionId=Xps6BhOnFjNsocp8O2sst0_8uktYVMPK"
-  echo "e879dbbdafb38fe6d99fb70d3297cfb9cd147c5e5cb527427bba9a929dd8d94c  bosh-agent" | shasum -a 256 -c -
+  curl -L -o bosh-agent "https://s3.amazonaws.com/bosh-agent-binaries/bosh-agent-2.29.0-linux-amd64?versionId=zKGvvFZ3edl4qqQL5QY7yKA0oJDLE6CY"
+  echo "33ef78e0d298e25d54f211fce15551fb40d0e09d7cd44d0ca5b8420d21fe9f64  bosh-agent" | shasum -a 256 -c -
```
dpb587-pivotal: Merger beware there will be a conflict here; ensure the latest version from master is correctly applied.

stemcell_builder/stages/untar_base_os_image/apply.sh
```diff
@@ -6,7 +6,7 @@ base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
 
 # the tar_base_os_image stage creates a tar ball of the chroot with no enclosing directory
-tar zxf $os_image_tgz -C $chroot
+sudo tar zxf $os_image_tgz -C $chroot
```
dpb587-pivotal: Why is this `sudo` necessary?

ericpromislow: I forget why. I'll look at this as well tomorrow.

stemcell_builder/stages/bosh_go_agent/apply.sh
ericpromislow: Have to go, I'll look at this Thursday AM

PR #10, opened 2017-06-26
Fix filename of stemcells downloaded with curl

Kiemes:
Currently we tell the user to download stemcells from bosh.io with this command:
```
curl -L -J -O https://bosh.io/d/stemcells/bosh-openstack-kvm-ubuntu-trusty-go_agent?v=3421.9
```
This leads to a file on our disk with the name `bosh-openstack-kvm-ubuntu-trusty-go_agent?v=3421.9` and not `bosh-stemcell-3421.9-openstack-kvm-ubuntu-7-go_agent.tgz` as it was intended. Clicking the download link on bosh.io will download with the correct filename. 

The `content-disposition` http header is used to set the filename when using
`curl -J -O` which at the moment was missing from our stemcells on S3.

[#147759327](https://www.pivotaltracker.com/story/show/147759327)

Signed-off-by: Mauro Morales <mamorales@suse.com>

### DIFF
--- a/ci/tasks/publish.sh
+++ b/ci/tasks/publish.sh
@@ -9,10 +9,11 @@ for file in $COPY_KEYS ; do
   file="${file/\%s/$VERSION}"
 
   echo "$file"
+  filename=$(${file} | grep -E -o 'bosh-stemcell.*.tgz' )
 
   # occasionally this fails for unexpected reasons; retry a few times
   for i in {1..4}; do
-    aws s3 cp "s3://$CANDIDATE_BUCKET_NAME/$file" "s3://$PUBLISHED_BUCKET_NAME/$file" \
+    aws s3 cp --content-disposition filename=${filename} --metadata-directive REPLACE "s3://$CANDIDATE_BUCKET_NAME/$file" "s3://$PUBLISHED_BUCKET_NAME/$file" \
       && break \
       || sleep 5
   done
--- a/ci/tasks/publish.sh
+++ b/ci/tasks/publish.sh
@@ -9,7 +9,7 @@ for file in $COPY_KEYS ; do
   file="${file/\%s/$VERSION}"
 
   echo "$file"
-  filename=$(${file} | grep -E -o 'bosh-stemcell.*.tgz' )
+  filename=$(basename "$file")
 
   # occasionally this fails for unexpected reasons; retry a few times
   for i in {1..4}; do


Kiemes:
This should fix it for the newly published stemcells. All others would need to be updated manually/script, could you do this since we don't have write access to the bucket?

cppforlife:
this shouldnt be fixed via s3 but rather by bosh.io redirect. ill take a look why it used to work.

Kiemes:
@cppforlife I guess you are right that something on bosh.io might have changed.

However, I do not see the reason for not having a valid content-disposition header for the "real source" on S3 directly? Could you elaborate what the downsides from your point of view could be?

ci/tasks/publish.sh
```diff
@@ -9,10 +9,11 @@ for file in $COPY_KEYS ; do
   file="${file/\%s/$VERSION}"
 
   echo "$file"
+  filename=$(${file} | grep -E -o 'bosh-stemcell.*.tgz' )
```
dpb587-pivotal: I don't believe this works as expected. This would try to execute `${file}`, which doesn't resolve and isn't executable. I think you meant `echo "${file}"`. But, I think it's simpler to do...

    filename=$(basename "$file")

ci/tasks/publish.sh
```diff
@@ -9,10 +9,11 @@ for file in $COPY_KEYS ; do
...
 
   # occasionally this fails for unexpected reasons; retry a few times
   for i in {1..4}; do
-    aws s3 cp "s3://$CANDIDATE_BUCKET_NAME/$file" "s3://$PUBLISHED_BUCKET_NAME/$file" \
+    aws s3 cp --content-disposition filename=${filename} --metadata-directive REPLACE "s3://$CANDIDATE_BUCKET_NAME/$file" "s3://$PUBLISHED_BUCKET_NAME/$file" \
```
dpb587-pivotal: This is setting the filename to the exact name of the object, so this is essentially a no-op. I don't think we should be duplicating that information.

Kiemes:
@cppforlife @dpb587-pivotal any update on the bosh.io issue? I fixed the filename just in case you might reconsider merging this ;)

voelzmo:
ping @cppforlife @dpb587-pivotal

dpb587-pivotal:
Seems like osx curl behavior is somewhat ambiguous -
 https://stackoverflow.com/questions/6881034/curl-to-grab-remote-filename-after-following-location#comment49197184_8841522

For better compatibility, maybe bosh.io should use `response-content-disposition`.

Kiemes:
ping ... @dpb587-pivotal @cppforlife 
We had a customer facing a similar issue.

dpb587-pivotal:
For people's sanity, it sounds like we should make this change, even though it's to workaround buggy client behavior and not a technically optimal reason.

voelzmo:
So, are we making this change? :)

dpb587-pivotal:
Oops. I clicked merge.

PR #14, opened 2017-07-19
Add second settings source to warden agent.json.

loewenstein:
Kubernetes only updates secrets/config mounted as directory.
In order to use a plain warden stemcell for the Kubernetes CPI
we need to hook in a settings source pointing to a file that will
be the only file in that folder.

### DIFF
--- a/stemcell_builder/stages/base_warden/apply.sh
+++ b/stemcell_builder/stages/base_warden/apply.sh
@@ -75,6 +75,10 @@ cat > $chroot/var/vcap/bosh/agent.json <<JSON
         {
           "Type": "File",
           "SettingsPath": "/var/vcap/bosh/warden-cpi-agent-env.json"
+        },
+        {
+          "Type": "File",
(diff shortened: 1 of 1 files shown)


loewenstein:
@cppforlife what about this PR?

cppforlife:
@loewenstein going to close this one for now. it appears that it's possible to use current warden stemcell configuration (as done by cfibmers/bosh-kubernetes-cpi). we'll eventually make that name better to decouple from particular name.

loewenstein:
@cppforlife the problem is that this way agent settings are going to be
immutable (a config map/secret mounted as single file will not be updated
when changed).

While the most prominent change is mounting disks and we anyway recreate
the pod for that, I think we should have updateable agent settings in
general as we do not want to have a separate registry.

If you say that disk operation is really going to be the only reason to
change agent settings though...
Is that the case now and in the foreseeable future?

Dmitriy Kalinin <notifications@github.com> schrieb am Fr., 19. Jan. 2018,
22:52:

PR #16, opened 2017-07-31
stemcell_builder: Include google_gcscli in Google stemcell

Everlag:
This change adds the google_gcscli tool to the Google stemcell to
support GCS as a native blobstore.

cc @evandbrown

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
@@ -141,6 +141,7 @@ module Bosh::Stemcell
 
     def google_stages
       [
+        :google_gcscli,
         :system_network,
         :system_google_modules,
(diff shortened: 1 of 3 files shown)


bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
```diff
@@ -141,6 +141,7 @@ def aws_stages
 
     def google_stages
       [
+        :google_gcscli,
```
cppforlife: let's include gcscli in all stemcells similar to s3cli.

evandbrown: @cppforlife would that be in `agent_stages`?

Everlag: Done.

PR #18, opened 2017-08-23
Add special stages for building softlayer stemcell

mattcui:
There are some special requirements for Softlayer stemcell, for examples, we need to enable root ssh login to allow CPI to configure network and persistent disk currently. The changes in this PR all are only specific to Softlayer stemcell, they don't impact the stemcells for other IaaSes, so I believe it's safe to merge into the master branch. The merge of this PR will make us much easier to build a fully automated end-2-end pipeline to keep step with the community to deliver and publish Softlayer stemcells on bosh.io. Please have a review. Thanks.

/cc @maximilien @cppforlife


maximilien:
@mattcui did you ensure that the list of [new packages](bosh-stemcell/spec/assets/dpkg-list-softlayer-ubuntu.txt) to be added for SL is the minimum set? In other words, nothing in the list is optional?

mattcui:
@maximilien Yes, I believe so, for bluemix stemcell, we have more changes needed, for community stemcell, they are should be minimal.

bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
```diff
@@ -192,6 +192,7 @@ def softlayer_stages
           :bosh_harden,
           :bosh_enable_password_authentication,
           :bosh_softlayer_agent_settings,
+          :bosh_config_root_ssh_login,
```
cppforlife: @mattcui i would love to have a discussion about how to remove password auth and replace it with ssh keys for softlayer stemcell. hopefully @maximilien can help us arrange a call?

maximilien: Can we plan this tomorrow Thursday September 28th, 2017 @ 5:30p PDT?

Beijing team comment here. We might also be able to do 6p. If yes, I'll schedule Zoom.

Best,

max

mattcui: @maximilien sorry that I have already on vacation and holiday since today, I will not be back until Oct. 9, so have to talk after the vacation.

mattcui:
We plan to move to new SL CPI (CPI NG) in Nov., the new CPI will not rely on VM configuration over ssh through user/password authentication. We will see if we still have special requirements for SL.

PR #19, opened 2017-09-14
Update Azure agent for CentOS stemcell issues

OpusDude:
The Azure CentOS 7 stemcells have been failing certification pipelines since April of 2017. It appears to be an issue with the Windows Azure Linux Agent [WALA] which is pretty old (v2.13). Building both the Ubuntu 14.04 and CentOS 7 stemcells with the newest WALA version (2.2.16) seems to resolve the issue and does not cause any regression.

### DIFF
--- a/stemcell_builder/stages/system_azure_wala/apply.sh
+++ b/stemcell_builder/stages/system_azure_wala/apply.sh
@@ -8,8 +8,8 @@ source $base_dir/lib/prelude_apply.bash
 packages="python python-pyasn1 python-setuptools"
 pkg_mgr install $packages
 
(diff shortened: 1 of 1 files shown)


tjvman:
Looks like [you already did this](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/b6e14d720dc90d5e0585c387efac04f0c978f401#diff-d20a49936e524a06552c85ba7ce311d9L11) as part of the xenial PR! 

Given that fact, seems like this PR should be closed. @dpb587-pivotal @OpusDude

PR #20, opened 2017-09-27
Ubuntu 16.04 (Xenial)

OpusDude:
This PR adds Ubuntu 16.04 (Xenial) as a stemcell option. It has been successfully built for all clouds but only verified on Azure. It "should" not break Ubuntu 14.04 or CentOS 7 builds. 

It does disable one test case in 'bosh-stemcell/spec/support/os_image_upstart_shared_examples.rb' which is for Ubuntu 14.04 only. I could not figure out an easy way to determine the DIST_CODENAME in rspec.


cppforlife:
@OpusDude a little update for pulling this in... we are currently in the middle of pulling in opensuse pr, and once that one is merged we are planning to pull in this pr. since bunch of people are at the summit, things are going a bit slow.

mxplusb:
@cppforlife just wanted to touch base on this issue to get an update. :)

/cc @mkocher @jcalvert1984

PR #23, opened 2017-11-30
Update SL Xenial stemell to get it work with SL CPI NG

edwardstudy:
Hi. We have already refactored SoftLayer CPI to improve stability and add new functionalities. 
More details info is in https://github.com/cloudfoundry/bosh-softlayer-cpi-release/tree/cpi_ng.

As new SoftLayer cpi switched to HTTP Registry to store VM user data, we also need to modify agent settings file to specify `userDataPath` and  `instanceIDPath` which make agents could read settings content successfully. So we changed `bosh_softlayer_agent_settings` stage. 

And CPI didn't need to configure VM networks and disks through SSH access, the agent would configure these settings. So we returned to disallow root login in stemcell builder and added softlayer-specified dpkg list file for testing.  

These updates make SoftLayer stemcell more consistent with the community and improve the automation of building SoftLayer stemcell CI.

BTW, the relative PR of bosh-agent is almost ready. It contains changing network setup and adding ISCSI device path resolution.

/cc @maximilien

### DIFF
--- a/bosh-stemcell/spec/support/os_image_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_shared_examples.rb
@@ -203,7 +203,7 @@ shared_examples_for 'every OS image' do
       expect(sshd_config).to be_mode(0600)
     end
 
-    it 'disallows root login (stig: V-38613)' do
+    it 'disallows root login (stig: V-38613)', exclude_on_softlayer: true do
       expect(sshd_config.content).to match(/^PermitRootLogin no$/)
     end
 
@@ -290,11 +290,11 @@ shared_examples_for 'every OS image' do
       expect(sshd_config.content).to match(/^Protocol 2$/)
       end
 
-    it 'sets AllowGroups to bosh_sshers (CIS 9.3.13)' do
+    it 'sets AllowGroups to bosh_sshers (CIS 9.3.13)', exclude_on_softlayer: true do
       expect(sshd_config.content).to match(/^AllowGroups bosh_sshers$/)
     end
 
-    it 'sets DenyUsers to root' do
+    it 'sets DenyUsers to root', exclude_on_softlayer: true do
       expect(sshd_config.content).to match(/^DenyUsers root$/)
     end
   end
--- a/stemcell_builder/stages/bosh_config_root_ssh_login/apply.sh
+++ b/stemcell_builder/stages/bosh_config_root_ssh_login/apply.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2009-2012 VMware, Inc.
+
+set -e
+
+base_dir=$(readlink -nf $(dirname $0)/../..)
+source $base_dir/lib/prelude_apply.bash
+
+sed "/^ *PermitRootLogin/d" -i $chroot/etc/ssh/sshd_config
+echo 'PermitRootLogin yes' >> $chroot/etc/ssh/sshd_config
+
+sed "/^ *AllowGroups/d" -i $chroot/etc/ssh/sshd_config
+
+sed "/^ *DenyUsers/d" -i $chroot/etc/ssh/sshd_config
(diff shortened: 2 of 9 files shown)


EdwardStudy:
Hi, @maximilien . The additionals(kpartx, multipath-tools, open-iscsi) is relatived with iscisi function in bosh-agent. bosh-agent need uses these packages to setup isicsi devices. I will change the file referring to other stemcells.

maximilien:
Sweet thanks for the update.

I am sure @cppforlife can help schedule this for the BOSH team to review and eventually merge.

Best,

Max

bosh-stemcell/spec/support/os_image_shared_examples.rb
```diff
@@ -442,14 +442,6 @@
         expect(sshd_config.content).to match(/^PasswordAuthentication no$/)
       end
     end
-
-    context 'unless: softlayer', {
```
cppforlife: @EdwardStudy looks like we deleted this test for everyone. we should bring it back since we do want to check that PermitRootLogin is set to no.

EdwardStudy: @cppforlife I just saw same case in: https://github.com/bluebosh/bosh-linux-stemcell-builder/blob/6a40fe1aa2e00671c92052dae507d8b519a3ddec/bosh-stemcell/spec/support/os_image_shared_examples.rb#L206-L208. And I saw our pipeline passed case above. Did I miss something?

EdwardStudy:
@dpb587-pivotal Hi, Danny. I just bring the test case back. Thank you.

EdwardStudy:
Hi, I just updated stig test to skip V-38613 for SoftLayer. SoftLayer uses the registry to store VM settings and permits root login which ensures bosh-cli deploy director with the registry.

EdwardStudy:
Hi, has anyone known "7 new issues in dependencies"?

cppforlife:
> Hi, has anyone known "7 new issues in dependencies"?

not quite sure what's that about. may be @dpb587-pivotal seen this.

cppforlife:
@EdwardStudy if we merge this, will this break existing stemcell lines? i suppose this is also blocked on agent merges that are going on.

EdwardStudy:
@cppforlife Yes. new SL CPI stemcell is not compatible with old SL CPI. But agent could be.

edwardstudy:
@cppforlife  I updated for SL xenial stemcell. Please help to review.

stemcell_builder/stages/static_libraries_config/assets/xenial_static_libraries_list.txt
```diff
@@ -75,3 +75,5 @@
 /usr/lib/x86_64-linux-gnu/liby.a
 /usr/lib/x86_64-linux-gnu/libz.a
 /usr/src/linux-headers-4.15/tools/objtool/libsubcmd.a
+/usr/local/ssl/lib/libcrypto.a
```
jfmyers9: These two libraries seemed to be added to the global list of `xenial_static_libraries`, but this applies to all stemcells that are being built. Can you explain the reason for this change?

stemcell_builder/stages/bosh_config_root_ssh_login/apply.sh
```diff
@@ -0,0 +1,17 @@
...
+source $base_dir/lib/prelude_apply.bash
+
+sed "/^ *PermitRootLogin/d" -i $chroot/etc/ssh/sshd_config
+echo 'PermitRootLogin yes' >> $chroot/etc/ssh/sshd_config
+
+sed "/^ *AllowGroups/d" -i $chroot/etc/ssh/sshd_config
+#echo 'AllowGroups bosh_sshers' >> $chroot/etc/ssh/sshd_config
+
+sed "/^ *DenyUsers/d" -i $chroot/etc/ssh/sshd_config
+#echo 'DenyUsers root' >> $chroot/etc/ssh/sshd_config
```
jfmyers9: Is there a reason to leave the commented out lines? Also can you please add a newline at the end of this file?

PR #26, opened 2018-01-16
rsyslogd should not query root zone regularly

wolfoo2931:
Each VM deployed by BOSH performs multiple DNS root queries every 30
seconds. It turned out that rsyslogd is performing these queries. In
larger system these queries sum up to a remarkable amount.

This change prevents rsyslogd to query the DNS root zone constantly.

### DIFF
--- a/stemcell_builder/stages/rsyslog_config/assets/rsyslog.conf
+++ b/stemcell_builder/stages/rsyslog_config/assets/rsyslog.conf
@@ -5,6 +5,8 @@
 #
 #  Default logging rules can be found in /etc/rsyslog.d/50-default.conf
 
+global(net.enableDNS="off")
+global(net.aclResolveHostname="off")
 
 #################
 #### MODULES ####


cppforlife:
@wolfoo2931 do you have any additional rsyslog configuration made to your VM (via syslog-release for example)? im not observing any traffic on an empty deployed VM.

wolfoo2931:
@cppforlife for some reasons it only happens on a Cloud Foundry VM (e.g. api (Cloud Controller)).

cppforlife:
@anexper did we ever find out if that setting only controller reverse dns?

anEXPer:
No, unfortunately we've not been able to get back to it yet.

We have a story to investigate here: https://www.pivotaltracker.com/story/show/154642150

anEXPer:
We tried these changes.

The `global(net.aclResolveHostname="off")` line appears to cause a segfault when rsyslog reads the config.

With that line commented out, rsyslog still seems able to forward logs to papertrail via hostname, which addresses our concern that disabling "DNS" might be broader than we want. We didn't test if reverse DNS lookups were still happening.

Have you tested these changes?

wolfoo2931:
Hey @anEXPer,

so far I remember, both lines where necessary to get rid of the reverse DNS lookups. But I'll try out again and let you know. 

I've tested these changes by inserting the lines into the configuration of an already running machine. No segfault appeared.

aashah:
This PR has been open for a while. I tried reproducing this on a more recent deployment of CF and was not able to do so. I'm gonna close this PR out, please feel free to comment if there's more context.

PR #34, opened 2018-03-22
Oracle Cloud stemcell

dmutreja:


aashah:
Hey @dmutreja,

The BOSH team is move away from adding further build targets into the bosh-linux-stemcell-builder. Recently, we've removed all targets but ubuntu-xenial. We recommend forking the repository instead.

I'm going to close this PR out, but let's please keep the conversation going if there's more context.

Thanks,
Aakash

PR #37, opened 2018-04-11
Fix for enabling XFS on ubuntu trusty stem cells on Google KVM 

babupe-g:
https://github.com/cloudfoundry/bosh/issues/1722

This issue is currently blocking multiple users from installing XFS filesystems on Ubuntu Trusty stem cells on Google Cloud VMs.
The only workaround for this has been to use persistent disks greater than 2TB. This solution might not be workable for all cases.

This change will make parted as the default FS partitioner on trusty images on Google Cloud.

### DIFF
--- a/stemcell_builder/lib/prelude_agent.bash
+++ b/stemcell_builder/lib/prelude_agent.bash
@@ -6,3 +6,11 @@ function get_partitioner_type_mapping {
       echo ''
   fi
 }
+
+function get_google_partitioner_type_mapping {
+  if [ "$(get_os_type)" == "opensuse" ] || ( [ "$(get_os_type)" == "ubuntu" ] && [ "${DISTRIB_CODENAME}" == "xenial" ]) || ( [ "$(get_os_type)" == "ubuntu" ] && [ "${DISTRIB_CODENAME}" == "trusty" ]); then
+      echo '"PartitionerType": "parted",'
+  else
+      echo ''
+  fi
+}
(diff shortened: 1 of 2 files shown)


babupe-g:
Ran the below test on the image:
cd /opt/bosh/bosh-stemcell; \
    STEMCELL_IMAGE=/mnt/stemcells/google/kvm/ubuntu/work/work/google-kvm-ubuntu.raw \
    STEMCELL_WORKDIR=/mnt/stemcells/google/kvm/ubuntu/work/work/chroot \
    OS_NAME=ubuntu \
    bundle exec rspec -fd --tag ~exclude_on_google \
    spec/os_image/ubuntu_trusty_spec.rb \
    spec/stemcells/ubuntu_trusty_spec.rb \
    spec/stemcells/go_agent_spec.rb \
    spec/stemcells/google_spec.rb \
    spec/stemcells/stig_spec.rb \
    spec/stemcells/cis_spec.rb > test_results.log

[test_results.log](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/files/1897210/test_results.log)

cppforlife:
@babupe-g we dont quite want to enable it for everyone. at the stemcell level configuration is global unfortunately. 

i've added https://www.pivotaltracker.com/story/show/156732961 to provide a runtime option for the agent. we'll try to do it shortly, but if you want to tackle it first, let us know.

PR #38, opened 2018-04-12
Support Azure serial console

gossion:
This PR is to enable serial console on hyper-v ubuntu stemcells.

[Serial console](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/serial-console) is a very good feature for diagnostics, with which users can still operate the VM even if they can't (bosh) ssh to the VM. Many people like this feature.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
@@ -173,6 +173,7 @@ module Bosh::Stemcell
       [
         :system_azure_network,
         :system_azure_wala,
+        :system_azure_serial_console,
         :system_parameters,
         :enable_udf_module,
         :bosh_clean,
--- a/bosh-stemcell/spec/bosh/stemcell/stage_collection_spec.rb
+++ b/bosh-stemcell/spec/bosh/stemcell/stage_collection_spec.rb
@@ -405,6 +405,7 @@ module Bosh::Stemcell
           [
             :system_azure_network,
             :system_azure_wala,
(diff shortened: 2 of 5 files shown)


cppforlife:
@gossion i would prefer if we only support it in ubuntu xenial+ which uses systemd. how urgent is this requirement? we are 2-3 weeks away from publishing xenials on bosh.io

gossion:
Thanks @cppforlife 

The requirement is not very urgent, as the feature of serial console is only in public preview on Azure.

We would like to see that users can do diagnostics in the further when they get issues like “provisioning timeout” or so in Ubuntu trusty and xenial. The feature is very helpful in issue investigations. Any reason that you don't want it in trusty?

cppforlife:
> Any reason that you don't want it in trusty?

im just hoping to add less new functionality to trusty, as we prepare for release xenial. 

i also see that bosh-stemcell/spec/stemcells/ubuntu_xenial_spec.rb was changed. does this actually work with systemd?

gossion:
Hi @cppforlife ,
I just confirmed that serial console is already supported on xenial stemcells on Azure. I added a commit [c4eba09](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/38/commits/c4eba09ec23a9d458ee17fcb8dd39665e6a163c8) to revert the change on it.

Does it mean xenial will be the main supported version gradually while trusty will be deprecated? if that is true, we should not add a feature to the retiring version, so we can close this PR directly.

gossion:
Close this PR, will not support serial console on Azure ubuntu trusty for now.
Thanks @cppforlife .

PR #46, opened 2018-05-25
Install golang 1.x from library image on SUSE

luan:
This normalizes the installation process with what we've been doing on the rest of bosh.

@manno @aduffeck could you take a look at this and tell us if you have any concerns before we merge it in?

CC @dpb587-pivotal


aashah:
We're planning to remove SUSE as part of an effort to reduce build targets. Going to close this issue and remove the underlying branch.

Mailing list: https://lists.cloudfoundry.org/g/cf-bosh/topic/upcoming_changes_to/65618897?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,65618897

PR #50, opened 2018-06-14
Wip zfs 157679109

BeckerMax:
Context in:
https://www.pivotaltracker.com/n/projects/956238


bosh-stemcell/spec/os_image/ubuntu_xenial_spec.rb
```diff
@@ -195,6 +195,17 @@
       it { should be_file }
       it { should be_executable }
     end
+
+    context 'zfs' do
+      %w[
+        /lib/modules/*/kernel/zfs/
+        /usr/src/linux-headers-*/zfs
+      ].each do |folder|
+        describe file(folder) do
```
dpb587-pivotal: Does `file` automatically do glob'ing, or is literally looking for a `*` in the path?

PR #52, opened 2018-06-26
Fix monit to recover from "Execution failed" if process is actually running

jandubois:
Patch originates from https://github.com/pivotal-cf/pcfdev-monit/compare/5c269d7...b398fcd

Background: We are doing some "external" monitoring that relies on the output of `monit summary` inside each VM/container.  Unfortunately this sometimes reports "Execution failed" when the process is actually running. I found a patch in the pivotal-cf/pcfdev-monit repo, that I think should be applied here as well.


ramonskie:
i noticed this pr and was wondering if it is still relevant

jandubois:
> i noticed this pr and was wondering if it is still relevant

I don't know. It might still be relevant, but if you are asking if we are still waiting for it to be merged, then the answer is "no". 😄 

So I'm going to close it; if somebody else cares, they can always re-open it (I'll leave the PR branch alone, at least for now).

PR #58, opened 2018-07-28
send the real hostname in dhcp on azure.

andyliuliming:


andyliuliming:
close the PR, because the option seems not necessary, wait for Pivotal guys to help confirm.

PR #59, opened 2018-08-01
disable the waagent auto update.

andyliuliming:


andyliuliming:
@mfine30 why we did this is because the waagent will be released time to time.
and the auto update is a bit dangerous for our stemcell. 
because the stemcell is a bit special than the standard gallary images. so it does not match the test matrix of waagent exactly.
so update the waagent manually and release the new stemcell is a more safer way, and a more fancy way why we use the stemcell.

Thanks.

mfine30:
Thanks @andyliuliming

PR #67, opened 2018-08-30
install the qemu-utils package for the stemcell for azure.

andyliuliming:
because we need config disk support for azure cpi.
and we need the qemu-utils to create a vhd.


andyliuliming:
close this temprily, to see whether we can make config disk work without qemu-img.

PR #68, opened 2018-09-03
add config disk setting for azure stemcell.

andyliuliming:
need add config disk setting in stemcell to enable the config disk scenario.


jfmyers9:
Hi @andyliuliming,

We don't have integration tests to validate that the Azure stemcells are successfully built in CI. Therefore we defer a lot of that validation to you all.

This change seems reasonable from a higher level. Have you validated that this configuration works on the Azure stemcell? If so, we can go ahead and merge the PR for you.

Best,

@jfmyers9 && @jrussett, CF BOSH

andyliuliming:
Hi @jfmyers9 yes, I validated it, and it works. thanks :)

PR #73, opened 2018-10-27
Make waagent depend on sshd.service

andyliuliming:
our WAAgent depends on the ssh host key,
so we need waagent wait for the sshd running to indicate the ssh host key generated succesfully.
so it's important to add
Before= to the waagent.service.


jaresty:
Merged manually (last commit is https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/77cc09f10477a51777a3c9a4169e08d27ed8d541)

PR #74, opened 2018-11-13
Remove expired certificate and "private" key

aeijdenberg:
Unclear to me what impact this might have - what I will note is that anything that accepted this expired SHA1 certificate and so-called private key is not receiving the confidentiality and integrity benefits that they might expect.

From digging by @beyhan, it appears this key/cert is used as a [fallback](https://github.com/cloudfoundry/bosh-agent/blob/e5514efd7b2388ef65980f1109dd7fecf752384c/mbus/https_dispatcher.go#L79-L89) when [mbus.cert](https://github.com/cloudfoundry/bosh-deployment/blob/13cf7e11fba28c98f81f339008d67cea4f5a95d0/bosh.yml#L21) is not defined, however if that actually succeeds, it would mean the client is misconfigured to allow expired certificates.

In any case, there doesn't seem any benefit in writing bogus unused credentials to the stemcells.


dpb587-pivotal:
Nice catch. I think this is a good thing to remove (we should also ensure that bosh-agent no longer even tries to read these files as well). Historically they were kept around to allow people to migrate their workflows, but all configurations should definitely be generating their own certificates by now.

@mfine30 this probably makes sense as a medium-low priority with benefits of removing some old, outdated code and making sure someone doesn't accidentally use an insecure key. Removing agent's support for looking at these default files should also be a separate story and part of this (I don't think there's a PR for that change).

PR #75, opened 2018-12-12
Don't duplicate log entries and remove unneeded log files

KaiHofstetter:
- as default log file use /var/log/syslog on Ubuntu and
  /var/log/messages on CentOS
- don't write debug, lpr.log, mail.*, user.log and news.* files.
  They are already written into the default log file. These
  log entries are unimportant and we don't need extra files for them.

[#161684006](https://www.pivotaltracker.com/story/show/161684006)

Co-authored-by: Kai Hofstetter <kai.hofstetter@sap.com>

### DIFF
--- a/stemcell_builder/stages/rsyslog_config/assets/rsyslog_50-default.conf
+++ b/stemcell_builder/stages/rsyslog_config/assets/rsyslog_50-default.conf
@@ -10,35 +10,35 @@ auth,authpriv.*			/var/log/auth.log
 #cron.*				/var/log/cron.log
 daemon.*			-/var/log/daemon.log
 kern.*				-/var/log/kern.log
-lpr.*				-/var/log/lpr.log
-mail.*				-/var/log/mail.log
-user.*				-/var/log/user.log
+#lpr.*				-/var/log/lpr.log
+#mail.*				-/var/log/mail.log
+#user.*				-/var/log/user.log
 
 #
 # Logging for the mail system.  Split it up so that
 # it is easy to write scripts to parse these files.
 #
-mail.info			-/var/log/mail.info
-mail.warn			-/var/log/mail.warn
-mail.err			/var/log/mail.err
+#mail.info			-/var/log/mail.info
+#mail.warn			-/var/log/mail.warn
+#mail.err			/var/log/mail.err
 
 #
 # Logging for INN news system.
 #
-news.crit			/var/log/news/news.crit
-news.err			/var/log/news/news.err
-news.notice			-/var/log/news/news.notice
+#news.crit			/var/log/news/news.crit
+#news.err			/var/log/news/news.err
+#news.notice		-/var/log/news/news.notice
 
 #
 # Some "catch-all" log files.
 #
-*.=debug;\
-	auth,authpriv.none;\
-	news.none;mail.none	-/var/log/debug
-*.=info;*.=notice;*.=warn;\
-	auth,authpriv.none;\
-	cron,daemon.none;\
-	mail,news.none		-/var/log/messages
+#*.=debug;\
+#	auth,authpriv.none;\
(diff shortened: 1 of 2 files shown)


johnpfield:
Our team has reviewed the PR and we think this is safe from an audit perspective.

The main concern is if there's any code in the CF ecosystem explicitly looking for logs to be in /var/log/debug, /var/log/messages, or /var/log/syslog.

We found one example where a garden test assumes it's output will end up in /var/log/syslog, which will break on CentOS:
https://github.com/cloudfoundry-incubator/garden-acceptance/blob/master/networking_test.go

Having different default locations on CentOS/Ubuntu could be a concern from an ecosystem-consistency POV, but stemcell-builder already has distro-specific changes (e.g. dev/console vs dev/xconsole)

stemcell_builder/stages/rsyslog_config/assets/rsyslog_50-default.conf
```diff
@@ -10,35 +10,35 @@ auth,authpriv.*			/var/log/auth.log
 #cron.*				/var/log/cron.log
 daemon.*			-/var/log/daemon.log
 kern.*				-/var/log/kern.log
-lpr.*				-/var/log/lpr.log
-mail.*				-/var/log/mail.log
-user.*				-/var/log/user.log
+#lpr.*				-/var/log/lpr.log
+#mail.*				-/var/log/mail.log
+#user.*				-/var/log/user.log
```
aashah: Could we delete these lines entirely from this file instead of leaving them commented?

Thanks,
Aakash & @jaresty

aashah:
Hey, thanks for the PR. Sorry for the delay in getting back to y'all.

Along the lines of what John mentioned above, have you validated whether this impacts common BOSH deployments (like cf-deployment or concourse). For example, have you tried deploying it and seeing whether it works, maybe CATs?

- Aakash & @jaresty

stemcell_builder/stages/rsyslog_config/assets/rsyslog_50-default.conf
NautiluX: I think we decided to keep them as this is how it looks like in the [default configuration for ubuntu](https://github.com/rsyslog/rsyslog-pkg-ubuntu/blob/99381a56fa1338d2d48c6e0aafd985226d8a7fb8/rsyslog/xenial/v8-stable/debian/50-default.conf#L9).

NautiluX:
I just executed CATS on BOSH-lite with these changes using the following config and they succeeded.
```
{
  "api": "api.bosh-lite.com",
  "apps_domain": "bosh-lite.com",
  "admin_user": "admin",
  "admin_password": "...",
  "artifacts_directory": "logs",
  "skip_ssl_validation": true,
  "timeout_scale": 1,
  "use_http": true,
  "use_log_cache": false,
  "stacks": [
    "cflinuxfs2",
    "cflinuxfs3"
  ]
}
```
Not sure if I should include any of the by default disabled tests, as only 82 tests are running with this configuration. Any thoughts @aashah?
```
Ran 82 of 213 Specs in 4142.906 seconds
SUCCESS! -- 82 Passed | 0 Failed | 1 Pending | 130 Skipped PASS
```

friegger:
@aashah, @jaresty

jaresty:
@friegger & @NautiluX I think we're fine with merging it based on these results -- keep an eye on [this story in the backlog](https://www.pivotaltracker.com/story/show/162607765) if you want to check where things are standing. We'll hopefully get to it soon; thanks for your patience.

stemcell_builder/stages/rsyslog_config/assets/rsyslog_50-default.conf
luan: Seems fair!

PR #76, opened 2018-12-16
add system_alicloud_modules

xiaozhu36:
The pr adds a system modules for alibaba cloud stemcell and aims to fix xenial stemcell bug.


aashah:
Nice and simple, will merge.

PR #80, opened 2019-02-07
Configure chrony to use hardware clock on Azure

ansd:
* On azure stemcells the sync-time script will be overwritten
* The chrony-updater-azure will configure chrony to use /dev/ptp0 as
hardware clock as described [in azure docs](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync#chrony)
* This prevents the need to configure UDP forwarding rules (necessary for NTP) in the Azure loadbalancer

Co-authored-by: David Ansari <david.ansari@sap.com>

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
@@ -202,6 +202,7 @@ module Bosh::Stemcell
         :system_azure_wala,
         :system_parameters,
         :enable_udf_module,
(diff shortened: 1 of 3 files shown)


dpb587-pivotal:
[Lack of] Historical context... I recall the Azure NTP with PTP suggestion having been discussed elsewhere at some point, but the only semi-related references I can find in Slack, email, and GitHub are https://github.com/cloudfoundry/bosh-agent/issues/182. From what I recall, it seemed like using PTP would be a good solution.

PR #81, opened 2019-02-08
Upgrade walinux agent to 2.2.36

dsboulder:
Microsoft and Pivotal field folks have discovered that some of the early
2.2.3x builds are imcompatible with azure stack, and requested we
upgrade to 2.2.35 or higher. Stevenson has validated
that the agent fails running on stack only on Xenial 170+, when we
upgraded the version of walinux agent.

Signed-off-by: David Stevenson <dstevenson@pivotal.io>


dsboulder:
We'd also love this in Xenial 250, and future releases

PR #83, opened 2019-03-13
Walinux rollback 2.2.25 for xenial 170

dsboulder:
Some mistakes were made when I asked you folks to upgrade to the latest walinux. We thought that would work with Azure Stack, because MS said they knew what the issue was, but it had the same problem as before. Since then I've built the stemcell locally and tried a lot of permutations of azure linux agent configurations. The only way I could get VMs to provision correctly was to rollback to the version of walinux we used in pre-170 xenial stemcells.

### DIFF
--- a/stemcell_builder/stages/system_azure_wala/apply.sh
+++ b/stemcell_builder/stages/system_azure_wala/apply.sh
@@ -8,8 +8,8 @@ source $base_dir/lib/prelude_apply.bash
 packages="python python-pyasn1 python-setuptools"
 pkg_mgr install $packages
 
-wala_release=2.2.36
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Is this supposed to be against 170.x, or master? Subject suggests to roll it back for just 170.x? An [old thread](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1543876034041800) was just resurrected which had originally blamed a WALA upgrade and wondering if this will be relevant to it.

dsboulder:
This issue has been seen in all stemcells 170+

dsboulder:
Also I don’t know how to do PRs well, so I could have messed up

PR #84, opened 2019-03-18
Allow logging with non-root users on xenial

s4heid:
This change lifts restrictions on the socket file `/dev/log`, which
enables non-root users to log to syslog. This is currently also the
default on basic Ubuntu Xenial OS images.

[fixes #164427238](https://www.pivotaltracker.com/story/show/164427238)

Co-authored-by: Manuel Dewald <m.dewald@sap.com>


aashah:
Hey,

We were about this merge this, and realized that there is _another_ group of people who are vested in security decisions. We have recently been trying to mindful of STIG requirements (think of it as a big checklist of security related items for distributing ubuntu images), that a compliance team has been actively been working on recently. We just reached out to them today, but we will merge if they give us the go ahead.

Thanks,
Aakash & @jaresty

Issue #90, opened 2019-04-18
[xenial] nvidia gpu support


alekssaul:
Hi,

I was interested to understand the possibility of using nvidia GPUs with CFCR/PKS however I had issues with getting nvidia drivers running on the Xenial stemcells. `nouveau` kernel modules seem to be enabled by default and runtime removal of the module seem to cause a panic on stemcell. 

I wanted to open a thread to discuss if it would be possible to blacklist `nouveau` 

+ @svrc

aashah:
Hello, closing this in favor of discussing the change in the PR.

Please feel free to re-open if you feel otherwise!

PR #96, opened 2019-06-25
blacklist nouveau kernel module

alekssaul:
Commit creates a `/etc/modprobe.d/blacklist-nouveau.conf` file that contains the modules required to disable the open source nouveau module.

`nouveau` module is useful to provide Open source support for nvidia GPUs and hardware acceleration for 3d rendering on desktops. On server side, users typically require the proprietary `nvidia` module which supports CUDA workloads for utilizing GPUs for compute capabilities. [Nvidia documentation](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#pre-installation-actions) provides this as a pre-requisite, along with the [thread](https://askubuntu.com/questions/841876/how-to-disable-nouveau-kernel-driver) from Ubuntu community. Swapping the kernel module from  `nouveau` to `nvidia` during runtime didn't seem possible as the BOSH resurrector replaced the VM based on my testing on Azure, when I forced removed the `nouveau` module on a running VM Instance.

I've tested AWS and Azure stemcell builds and I was able to use the stemcell image to build a PKS cluster with nvidia GPU attached workers and Tensorflow based workload using GPUs from Kubernetes.

This PR should address issue #90 .

### DIFF
--- a/stemcell_builder/stages/system_kernel_modules/apply.sh
+++ b/stemcell_builder/stages/system_kernel_modules/apply.sh
@@ -21,6 +21,13 @@ install udf /bin/true
 install rds /bin/true
 options ipv6 disable=1' >> $chroot/etc/modprobe.d/blacklist.conf
 
+echo '# prevent nouveau from loading
+blacklist nouveau
+blacklist lbm-nouveau
+options nouveau modeset=0
+alias nouveau off
+alias lbm-nouveau off' >> $chroot/etc/modprobe.d/blacklist-nouveau.conf
+
 if [[ $OS_TYPE == 'ubuntu' ]]; then
   rm -rf $chroot/lib/modules/*/kernel/zfs $chroot/usr/src/linux-headers-*/zfs
 fi


aashah:
Hey, just looking to check my understanding of a `modprobe.d/*.conf` file.

My understanding is the most holistic way of blacklisting a module was to insert a `blacklist <module name>`, as this overrides what the kernel will run when trying to load it.

Does the combination of `alias <module> off` and `blacklist <module>` achieve the same effect?

alekssaul:
Hi @aashah ,

I imagine `alias <module> off` bits can be removed since on a stemcell it's very unlikely (impossible?) for the modules to be aliased. I can make the change to remove the `alias` lines.

aashah:
Sure..what about the question around `install <module> /bin/true`?

alekssaul:
I was generally following the; `https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#runfile-nouveau-ubuntu` documentation. 
I can the commit to just include 
```
install nouveau /bin/true
install lbm-nouveau /bin/true
options nouveau modeset=0
```
as an append to $chroot/etc/modprobe.d/blacklist.conf to keep the change more consistent.

PR #110, opened 2020-02-28
[#168964744] make rsyslog and logrotate configs consistent

xtreme-nitin-ravindran:
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/75 removed duplicate/unneeded log file entries from `rsyslog_50-default.conf`, but these logs are still configured to be log-rotated.
This causes CIS level 1 Rule 4.3 to fail

### DIFF
--- a/bosh-stemcell/spec/support/os_image_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_shared_examples.rb
@@ -142,11 +142,6 @@ shared_examples_for 'every OS image' do
         expect(subject.content).not_to match 'restart rsyslog'
       end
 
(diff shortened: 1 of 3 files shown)


xtreme-nitin-ravindran:
@h4xnoodle : Can this change be back-ported to the existing xenial stemcell lines? My bad, I should have added the back port request when I opened the PR.

h4xnoodle:
done! backported down to 97.x. Some other little backports were required to make these ones work.

xtreme-nitin-ravindran:
Thanks. @h4xnoodle

PR #111, opened 2020-02-29
[#171553414] add CIS, STIG hardening for ssh

xtreme-nitin-ravindran:


xtreme-nitin-ravindran:
@h4xnoodle : Can this change be back-ported to the existing xenial stemcell lines? My bad, I should have added the back port request when I opened the PR.

PR #113, opened 2020-04-23
Add rng-tools to ubuntu based stemcells

metron2:
I noticed you guys already have this package for centos, but somehow it got left out of the ubuntu images.

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
@@ -403,6 +403,7 @@ python3.6-minimal
 quota
 readline-common
 resolvconf
+rng-tools
 rsync
 rsyslog
 rsyslog-gnutls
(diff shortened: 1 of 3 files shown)


mrosecrance:
The change suggested is only changing a test asset. The change needs to occur potentially here: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/bc4229752403476fd1444ec6fff624f3179f9436/stemcell_builder/stages/base_ubuntu_packages/apply.sh to be globally applied. I believe it will by default be enabled once installed there, so I'm also unsure if we need to add the --ignore flag now, or if we can wait until we see if users will report fips errors. I'll leave that bit up to you.

metron2:
Ok I've fixed that up.

The ignore flag isn't needed if its installed as part of ubuntu (it just gets restarted when it exits)

dlresende:
@metron2 Thanks for this PR.

Would you mind submitting another PR to the `ubuntu-bionic/master` branch?

Issue #115, opened 2020-05-22
Consider disabling port forwarding by default


friegger:
There are recommendations to turn off port forwarding by default, e.g. https://www.ssh.com/ssh/sshd_config/.

```
AllowTcpForwarding no
AllowStreamLocalForwarding no
GatewayPorts no
PermitTunnel no
```

Since the stemcell is a general purpose image it might be necessary to change this for certain scenarios. I assume that os-conf-release could learn to provide that possibility.

PR #117, opened 2020-06-22
Disable port forwarding by default on stemcells

krutten:
Issue #115 
 
[172972421](https://www.pivotaltracker.com/story/show/172972421)


friegger:
@krutten Since you provided the change, do you know who could check the release pipelines and cut a release? Have you considered the comment I made in the issue regarding whether this might break existing scenarios?

mrosecrance:
@krutten Have you tested that os-conf can re-enable port-forwarding?

friegger:
Hi all,

great that `os-conf-release` will now offer to lock this down. Could one of you elaborate a little how you came to the decision for that solution? I assume changing the defaults on the stemcell seemed too risky? Would cutting a new major version would be a solution?

dlresende:
Hi @friegger,
Changing the defaults on the stemcell could affect features such as SSH Access and Tunneling.
We are then proposing a change on `os-conf-release` to achieve the same intent.

Issue #114, opened 2020-05-05
Recommended nvme timeout not configered in AWS stemcell


friegger:
AWS recommends to set a very high NVMe timeout:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html

The stemcell runs with the default timeout of 30:
```
$ cat /sys/module/nvme_core/parameters/io_timeout
30
# Tested on a m5d.large instance with bosh-aws-xen-hvm-ubuntu-xenial-go_agent/621.64
```

friegger:
Adding slack conversation for completeness: https://cloudfoundry.slack.com/archives/C094QRLEQ/p1588694931011500

friegger:
Implemented for Bionic.

Issue #99, opened 2019-08-01
aws stemcell - limited network connectivity due to network driver restarts


BeckerMax:
**Symptoms:**
On AWS on the affected VMs we observe problems with the network connectivity. Frequently the network connectivity is down, coming back up shortly after. This is probably related to a restart of the NIC as seen in the logs below. It is independent from the releases running on the managed bosh vm.

The problem will stay indefinitely. To solve it we delete the vm via bosh. We once rebooted a VM which also solved it. 

**Logs:**
kern.log:
```
2019-07-22T11:01:55.096109+00:00 localhost kernel: [2008934.869922] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:06.108096+00:00 localhost kernel: [2008945.880952] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:11.992109+00:00 localhost kernel: [2008951.765109] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:21.972103+00:00 localhost kernel: [2008961.745862] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:32.216097+00:00 localhost kernel: [2008971.989764] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:42.964105+00:00 localhost kernel: [2008982.737861] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:53.208095+00:00 localhost kernel: [2008992.981733] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:02:59.100107+00:00 localhost kernel: [2008998.873855] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
2019-07-22T11:03:10.104101+00:00 localhost kernel: [2009009.877387] ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
(...)
```
There are no correlating log entries in the agent or audit.log.

**Impact**
Network calls will occasionally not go through. Usually the VM will report as healthy as the occasional health check goes through. Jobs will still be reported as `running`. VMs that rely on frequent network communication like Diego cells are strongly affected.

As the problem will stay indefinitely we currently need to spot the instances and then use `delete-vm` to delete them manually.

**Reproducibility**
[...]

PR #118, opened 2020-06-22
Increase the NVMe timeout (for EC2)

krutten:
[172676505](https://www.pivotaltracker.com/story/show/172676505)

### DIFF
--- a/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
@@ -15,6 +15,7 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
       its(:content) { should match ' cgroup_enable=memory swapaccount=1' }
       its(:content) { should match ' console=ttyS0,115200n8' }
       its(:content) { should match ' earlyprintk=ttyS0 rootdelay=300' }
+      its(:content) { should match ' nvme_core.io_timeout=255 nvme_core.max_retries=10' }
       its(:content) { should match %r{initrd\t/boot/initrd.img-\S+-generic} }
 
       it('should set the grub menu password (stig: V-38585)') { expect(subject.content).to match /^password --md5 \*/ }
--- a/bosh-stemcell/spec/stemcells/ubuntu_xenial_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_xenial_spec.rb
@@ -15,6 +15,7 @@ describe 'Ubuntu 16.04 stemcell image', stemcell_image: true do
       its(:content) { should match ' cgroup_enable=memory swapaccount=1' }
       its(:content) { should match ' console=ttyS0,115200n8 console=tty0' }
       its(:content) { should match ' earlyprintk=ttyS0 rootdelay=300' }
+      its(:content) { should match ' nvme_core.io_timeout=255 nvme_core.max_retries=10' }
       its(:content) { should match %r{initrd /boot/initrd.img-\S+-generic} }
 
       it('should set the grub menu password (stig: V-38585)') { expect(subject.content).to match /^password --md5 \*/ }
--- a/stemcell_builder/stages/image_install_grub/apply.sh
+++ b/stemcell_builder/stages/image_install_grub/apply.sh
@@ -86,7 +86,7 @@ default=0
 timeout=1
 title ${os_name} (${kernel_version})
   root (hd0,0)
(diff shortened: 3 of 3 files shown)


dlresende:
Addresses #114.

krutten:
Validated on AWS m5.large Nitro Instance

```
Old Stemcell:

emptyvm/3fbf3909-55b3-4755-a14e-ce8834257a26:~$ sudo cat /sys/module/nvme_core/parameters/io_timeout
30
emptyvm/3fbf3909-55b3-4755-a14e-ce8834257a26:~$ uname -a
Linux ad95db65-dfef-4a74-b6ee-41a1d06813e0 4.15.0-112-generic #113~16.04.1-Ubuntu SMP Fri Jul 10 04:37:08 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
emptyvm/3fbf3909-55b3-4755-a14e-ce8834257a26:~$



New Stemcell:

emptyvm/12107834-eea1-41e3-b8b7-6a6b256e1c8a:~$ sudo cat /sys/module/nvme_core/parameters/io_timeout
255
emptyvm/12107834-eea1-41e3-b8b7-6a6b256e1c8a:~$
```

Also verified on vSphere and GCP and AWS non-Nitro as ignored if no NVMe present

friegger:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#timeout-nvme-ebs-volumes mentions that for current kernels the timeout should be configured to `4294967295`, but older kernels only support `255`. Does your change mean that the kernel used for the stemcell is such an older kernel? Which kernel version supports higher values?

dlresende:
@friegger we are configuring the timeout in a way it will be available across cloud providers. We've seen the AWS recommendation, but for deployments in other platforms this might have implications we are not sure of. We couldn't find a more generic recommendation so in that sense we are leaning towards increasing the timeout but not by much. I'd be curious if someone had other pointers that could shed more light around a consensual value.

Kernel versions should not matter as the current supported stemcell versions could cope with bigger values.

voelzmo:
@dlresende out of curiosity: which other IaaS layers use NVMe except for the AWS nitro instance types? It feels weird to have a recommendation by a cloud provider (most likely for a reason) and not follow it. This most likely causes issues in case you'd need storage related support, we've seen this with network related issues as well.

krutten:
Hi @voelzmo, the original timeout was 30s.  We looked at the recommended timeout of 4294967295 and it is about 136 years, effectively forever.  Since it's hard to simulate nvme failures, we don't know what type of issues could happen since the instance use the volume for both packages and swap space.  The 255 is about four and a half minutes, which seemed a more reasonable timeout.  If there is an issue with the volume, it is now possible we will see a timeout and BOSH may be able to handle it better.  Now that the timeout has been bumped, we can see if there are any issues, and if needed increase it further in the future.

voelzmo:
Hey @krutten, thanks for the reasoning behind this change. 

My understanding of the AWS recommendations is similar to yours: they're suggesting to set it to "forever", depending on which kernel you're on this means either 255 or 4294967295. I'm not sure why they're suggesting this and in which cases a value bigger than 255 seconds is necessary or even helpful – if we're interested in their reasons, I guess we should reach out to their support to help us understand.

In our experience with AWS support, they're pretty hard to argue with until you follow their instructions by the book. This is what happened for example in the context of #99: They wouldn't help us much unless we did everything as described in their instructions for the network driver and even asked us to use the awslinux kernel.

The big turnaround times on any change in the stemcell builder make iterating on an issue pretty cumbersome, I'd therefore like to use the settings AWS recommends or find a reason why this doesn't make sense for us. You're right in the sense that probably a setting like 255 doesn't make that big of a difference (as they don't put any caveat regarding older kernels and possible issues on the NVMe page), but this seems like a magic number we're never going to touch anymore and nobody can explain in 2 months from now.

You mentioned that BOSH could handle a timeout, and I'd like to understand more about this. To the best of my knowledge, one thing that BOSH _doesn't_ particularly cope well with is storage failures, e.g. a disk being mounted read-only or similar. In the past, this occurred a few times for us on OpenStack and people ended up writing their own monitoring and alerting for these cases. Do you have anything specific in mind here, where an NVMe related kernel timeout could help us propagate an error to BOSH?

MSSedusch:
@krutten is it possible to set this timeout only for AWS stemcells? For Azure, we have a VM type (Lsv2) that has local NVMe drives and in this scenario we recommend keeping the default.

jpalermo:
Resolved in #140

aramprice:
@krutten - is it ok to delete this branch?

PR #140, opened 2021-01-13
Increase nvme io timeout for AWS for bionic

StefanWutz:
This PR increases the nvme io timeout on (AWS https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/114) for bionic stemcells. Other cloud provider still using the default value 30 seconds defined in the linux kernel (https://github.com/torvalds/linux/blob/master/drivers/nvme/host/core.c#L38). This implementation considers the discussion in PR https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/118.

### DIFF
--- a/stemcell_builder/stages/image_aws_install_grub/apply.sh
+++ b/stemcell_builder/stages/image_aws_install_grub/apply.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+
+set -e
+
+base_dir=$(readlink -nf $(dirname $0)/../..)
+source $base_dir/lib/prelude_apply.bash
+
+disk_image=${work}/${stemcell_image_name}
+image_mount_point=${work}/mnt
+
+## unmap the loop device in case it's already mapped
+#umount ${image_mount_point}/proc || true
+#umount ${image_mount_point}/sys || true
+#umount ${image_mount_point} || true
+#losetup -j ${disk_image} | cut -d ':' -f 1 | xargs --no-run-if-empty losetup -d
+kpartx -dv ${disk_image}
+
+# note: if the above kpartx command fails, it's probably because the loopback device needs to be unmapped.
+# in that case, try this: sudo dmsetup remove loop0p1
+
+# Map partition in image to loopback
+device=$(losetup --show --find ${disk_image})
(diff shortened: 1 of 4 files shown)


stemcell_builder/stages/image_aws_install_grub/apply.sh
```diff
@@ -0,0 +1,113 @@
...
+echo "(hd0) ${device}" > ${image_mount_point}/device.map
+
+# install bootsector into disk image file
+run_in_chroot ${image_mount_point} "grub-install -v --no-floppy --grub-mkdevicemap=/device.map --target=i386-pc ${device}"
+
+# Enable password-less booting in openSUSE, only editing the boot menu needs to be restricted
+run_in_chroot ${image_mount_point} "sed -i 's/CLASS=\\\"--class gnu-linux --class gnu --class os\\\"/CLASS=\\\"--class gnu-linux --class gnu --class os --unrestricted\\\"/' /etc/grub.d/10_linux"
+cat >${image_mount_point}/etc/default/grub <<EOF
+GRUB_CMDLINE_LINUX="vconsole.keymap=us net.ifnames=0 biosdevname=0 crashkernel=auto selinux=0 plymouth.enable=0 console=ttyS0,115200n8 earlyprintk=ttyS0 rootdelay=300 ipv6.disable=1 audit=1 cgroup_enable=memory swapaccount=1 nvme_core.io_timeout=4294967295"
+EOF
```
mrosecrance, line 78: It seems unideal to copy this whole file to add a param, especially if we want to make more grub config changes in the future. Have you all tried appending the timeout param as a seperate ops file that runs next? It looks like the rest of this file is just setting up config.

StefanWutz: Indeed this was my first intention just to change the `GRUB_COMMAND_LINE`, call `grub-mkconfig` and place the script file after `image_install_grub`. The problem to me is that there will be some more changes on the generated files afterwards and I need to overtake more lines of code which makes it at the end confusing. So I decided to replaced the whole file for AWS. To introduce a new variable for the cloud provider seems also not to fit to the code structure.

PR #141, opened 2021-02-03
Remove ext4 feature metadata_csum for alicloud

StefanWutz:
The ext4 feature 'metadata_csum' is by default enabled on bionic. Alicloud does not support this feature at the moment. VMs based on images containing an ext4 partition formatted with 'metadata_csum' will not start.

This PR will remove the feature from file `/etc/mke2fs.conf` inside the docker container so ext4 partitions created during the build process do not contain this feature. Additional the feature is disabled inside the image so that created ext4 partitions on a running vm do not contain this feature as well.

Please note that the problem does not occur with alicloud stemcell version 0.6 due it was build with an xenial docker container which does not support this ext4 feature.


stemcell_builder/stages/system_alicloud/apply.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+base_dir=$(readlink -nf $(dirname $0)/../..)
+source $base_dir/lib/prelude_apply.bash
+
+if [[ "${DISTRIB_CODENAME}" != 'xenial' ]]; then
+  echo "acpiphp" >> $chroot/etc/modules
```
mariash, line 9: Why is this module being loaded?

StefanWutz: This change was introduced with commit https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/cb71dbaab858085cb9fc9c71dcbde95f8ebe18fa.
I guess it was related to ubuntu trusty. On bionic it is loaded by default
```
> grep "CONFIG_HOTPLUG_PCI_ACPI=" /boot/config-`uname -r`
CONFIG_HOTPLUG_PCI_ACPI=y
```
Interestingly the code snippet does also exist for aws and openstack:
* https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-bionic/master/stemcell_builder/stages/system_aws_modules/apply.sh#L8-L10
* https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-bionic/master/stemcell_builder/stages/system_openstack_modules/apply.sh#L8-L10

I think it can be removed. What is your opinion?

voelzmo:
There's also this commit: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/741f485675f13ec3cda19d375b8f30b1aa1c584c

Seems like your assessment is absolutely correct, @StefanWutz, it seems Trusty related. This ``if`` probably should have been an ``if "${DISTRIB_CODENAME}" == 'trusty'``. But who knows ;)

stemcell_builder/stages/system_alicloud/apply.sh
mariash: It looks like this module was introduced to fix an issue in Ubuntu Trusty. So we can remove it from xenial and bionic. We will merge PR and clean up this.

Issue #132, opened 2020-11-15
stemcell ci test failed by step `disables ipv6 in the kernel`


xiaozhu36:
When I running the stemcell CI test and I happened the following error:
```
/tmp/build/87eb1201/bosh-linux-stemcell-builder/src/github.com/cloudfoundry/stemcell-acceptance-tests/ipv4director/smoke/smoke_test.go:73
------------------------------
[Cmd Runner] 2020/11/15 09:58:53 DEBUG - Running command '/usr/local/bin/bosh -n -d stemcell-acceptance-tests --column=stdout ssh default/0 -r -c sudo netstat -lnp | grep sshd | awk '{ print $4 }''
[Cmd Runner] 2020/11/15 09:59:03 DEBUG - Stdout: 0.0.0.0:22	
:::22	
	
[Cmd Runner] 2020/11/15 09:59:03 DEBUG - Stderr: 
[Cmd Runner] 2020/11/15 09:59:03 DEBUG - Successful: true (0)
• Failure [9.392 seconds]
Stemcell [It] #141987897: disables ipv6 in the kernel 
/tmp/build/87eb1201/bosh-linux-stemcell-builder/src/github.com/cloudfoundry/stemcell-acceptance-tests/ipv4director/smoke/smoke_test.go:92

  Expected
      <[]string | len:2, cap:2>: ["0.0.0.0:22\t", ":::22"]
  to equal
      <[]string | len:1, cap:1>: ["0.0.0.0:22"]
```
Is there anyone can help me?

StefanWutz:
I had a look on this issue and it seems that ipv6 is still enabled for vms running bionic stemcells on alicloud.

An ipv6 address is shown for the default nic

```
web/6e626a8c-35d4-467e-b297-26ade1c4467d:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:16:3e:00:d9:61 brd ff:ff:ff:ff:ff:ff
    inet 10.2.6.0/16 brd 10.2.255.255 scope global eth0
[...]

PR #142, opened 2021-02-23
WIP: Fix IPv6 issue in grub configuration on alicloud

StefanWutz:
This PR is a workaround to fix the IPv6 issues on alicloud as described in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/132.

**Problem:**
IPv6 is disabled `ipv6.disable=1` using the kernel boot parameters defined in the grub configuration `/etc/default/grub`. However on alicloud the kernel boot parameter `ipv6.disable=1` disappeared in files `/etc/default/grub` and `/boot/grub/grub.cfg` after a vm was created. Obviously _bosh-agent_ removed the parameter from configuration, but this is not the case as it only operate on file `/boot/grub/grub.cfg`. The logs are not showing any process which is manipulating the grub configuration on startup.
 
The file `/etc/default/grub`should never be changed on a vm, so we should expect:
```
GRUB_CMDLINE_LINUX="vconsole.keymap=us net.ifnames=0 biosdevname=0 crashkernel=auto selinux=0 plymouth.enable=0 console=ttyS0,115200n8 earlyprintk=ttyS0 rootdelay=300 audit=1 ipv6.disable=1 cgroup_enable=memory swapaccount=1"
```
but we found:
```
GRUB_CMDLINE_LINUX="vconsole.keymap=us net.ifnames=0 biosdevname=0 crashkernel=auto selinux=0 plymouth.enable=0 console=ttyS0,115200n8 earlyprintk=ttyS0 rootdelay=300 audit=1  cgroup_enable=memory swapaccount=1  console=tty0"
```

The root cause is still unknown.

**Workaround:**
For alicloud we add a new configuration item GRUB_CMDLINE_LINUX_DEFAULT to the grub configuration specifying again the kernel boot parameter `ipv6.disable=1`. This change will survive the boot process and IPv6 on alicloud is disabled. Enable IPv6 using the _bosh-agent_ works as expected (tested).

### DIFF
--- a/bosh-stemcell/spec/stemcells/alicloud_spec.rb
+++ b/bosh-stemcell/spec/stemcells/alicloud_spec.rb
@@ -24,4 +24,18 @@ describe 'AliCloud Stemcell', stemcell_image: true do
       its(:content) { should_not match /metadata_csum/ }
     end
   end
+
+  context 'default grub configuration' do
+    describe file('/etc/default/grub') do
+      it { should be_file }
+      its(:content) { should match '^GRUB_CMDLINE_LINUX_DEFAULT="ipv6.disable=1"' }
+    end
+  end
+
+  context 'installed by image_install_grub' do
+    describe file('/boot/grub/grub.cfg') do
+      it { should be_file }
+      its(:content) { should match ' ipv6.disable=1' }
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


StefanWutz:
DO NOT MERGE.

According to alicloud the Bionic stemcell image will be manipulated during image upload and the kernel boot parameter `ipv6.disable` will be removed. This change will not to Xenial images.

We are validating if this change can be removed or the correct behavior can be applied to the CPI.

StefanWutz:
The kernel boot parameters will be changed when uploading an Ubuntu Bionic image to alicloud. This includes the deletion of the parameter `ipv6.disable=1`. The modification is based on the Linux type 'Ubuntu'. Changing the Linux type to 'Other' will not lead to changed kernel boot parameters.
As discussed with alicloud team this will be fixed in the alicloud cpi, see https://github.com/cloudfoundry-incubator/bosh-alicloud-cpi-release/issues/122.

PR #152, opened 2021-04-07
do not use /var/vcap/store in the ns for chrony as chrony is configured to use a Private namespace see #151

### LINKED ISSUE #151: Detaching volume failed on AWS after upgrading to Bionic
After upgrading to Bionic recreation of vms using persistent disks are failing regular with error message

```
Task 118306 | 16:43:44 | Preparing deployment: Preparing deployment (00:00:07)
Task 118306 | 16:43:51 | Preparing deployment: Rendering templates (00:00:03)
Task 118306 | 16:43:54 | Preparing package compilation: Finding packages to compile (00:00:00)
Task 118306 | 16:43:55 | Updating instance bosh: bosh/d554ed4d-a459-43bb-afb1-be23ffc84be6 (0) (canary)
Task 118306 | 16:43:55 | L executing pre-stop: bosh/d554ed4d-a459-43bb-afb1-be23ffc84be6 (0) (canary)
Task 118306 | 16:43:55 | L executing drain: bosh/d554ed4d-a459-43bb-afb1-be23ffc84be6 (0) (canary)
Task 118306 | 16:44:04 | L stopping jobs: bosh/d554ed4d-a459-43bb-afb1-be23ffc84be6 (0) (canary)
Task 118306 | 16:44:05 | L executing post-stop: bosh/d554ed4d-a459-43bb-afb1-be23ffc84be6 (0) (canary) (00:27:01)
                      L Error: CPI error 'Bosh::Clouds::CloudError' with message 'Timed out waiting for volume vol-0c1c5e82d5b564f95 to be detached to instance i-06c7d7363eb1d6028 as device /dev/sdf to be detached, took 1604.50443333s' in 'detach_disk' CPI method (CPI request ID: 'cpi-190690')
Task 118306 | 17:10:56 | Error: CPI error 'Bosh::Clouds::CloudError' with message 'Timed out waiting for volume vol-0c1c5e82d5b564f95 to be detached to instance i-06c7d7363eb1d6028 as device /dev/sdf to be detached, took 1604.50443333s' in 'detach_disk' CPI method (CPI request ID: 'cpi-190690')Updating deployment:
  Expected task '118306' to succeed but state is 'error'
```

After some investigation we have found that the `chrony` service still holds a reference to the `/var/vcap/store` mount point. A restart `systemctl restart chrony.service` will fix this issue temporary and AWS can detach the volume.

The `chrony` service is running in it's own namespace. When the persistent disk has been mounted before the `chrony` service has been started or restarted, it will be referenced by the `chrony` service.

Systemd settings for `chrony` service which might related to this issue:

```
PrivateTmp=yes
ProtectHome=yes
ProtectSystem=full
```
We have seen this issue only on AWS so far.

ramonskie:
this PR will fix #151 
chrony uses  the following systemd settings 
```
PrivateTmp=yes
ProtectHome=yes
ProtectSystem=full
```
this results in a private namespace.
when `/var/vcap/store` is mounted and chrony restart.
the namespace created will then retrieve all the current active mountpoints.

if bosh will recreate the vm and wants to detach the disk. it will get stuck as chrony's namespace is still using the mountpoint
and mount does not have access to the namespace.

this pr will prevent the use that the ns will never mount  `/var/vcap/store`

### DIFF
--- a/stemcell_builder/stages/bosh_ntp/assets/restart.conf
+++ b/stemcell_builder/stages/bosh_ntp/assets/restart.conf
@@ -4,4 +4,4 @@ PIDFile=/var/run/chronyd.pid
 RemainAfterExit=no
 Restart=always
 RestartSec=5s
-
+InaccessiblePaths=/var/vcap/store
--- a/bosh-stemcell/spec/support/os_image_systemd_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_systemd_shared_examples.rb
@@ -15,6 +15,7 @@ shared_examples_for 'a systemd-based OS image' do
     describe file('/etc/systemd/system/chrony.service.d/restart.conf') do
       it { should be_file }
       its(:content) { should match /^Restart=always/ }
+      its(:content) { should match /^InaccessiblePaths=\/var\/vcap\/store/ }
     end
 
     describe file('/etc/systemd/system/var-log.mount.d/start_rsyslog_on_mount.conf') do


stemcell_builder/stages/bosh_ntp/assets/restart.conf
```diff
@@ -4,4 +4,4 @@ PIDFile=/var/run/chronyd.pid
 RemainAfterExit=no
 Restart=always
 RestartSec=5s
-
+InaccessiblePaths=/var/vcap/store
```
StefanWutz, line 7: Should we add a `-` prefix in case when the persistent disk does not exist?

See https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=
 
_"Paths in ReadWritePaths=, ReadOnlyPaths=, InaccessiblePaths=, ExecPaths= and NoExecPaths= may be prefixed with "-", in which case they will be ignored when they do not exist."_

ramonskie: i just tested all scenarios.
and when i add a `-` /var/vcap/store is being added again.

without the prefix both mounted and umounted works as designed

rkoster: @ramonskie does chrony also start on instances without a persistent disk?

ramonskie: @rkoster yes i tested the following.
- with persistent disk +prefix - var/vcap/store shows up in the ns again
- without persistent disk +prefix

- with persistent disk +without the prefix (current) - var/vcap/store does not shows up in the ns again
- without persistent disk +without the prefix (current)

ramonskie: a second quicktest. i noticed a flaw in my test.
and the prefix is needed.
i will fix that in another pr

PR #156, opened 2021-04-16
untill we switch to gomod

ramonskie:
set gomodule=off until we switch to gomod 
see failing test https://main.bosh-ci.cf-app.com/teams/main/pipelines/bosh:stemcells:ubuntu-bionic/jobs/test-stemcells-0.x-ipv4/builds/17


mrosecrance:
I'm starting the work to just migrate the 621 branch to go mod. I haven't looked at what might have cause the docker image to bump to go 1.16 (which default go mod to on) but we have only until go 1.17 where that won't be an option anymore.

PR #175, opened 2021-06-24
build stemcell for CloudStack with Xen

romain-dartigues:
This patch make it possible to build a stemcell to be used on the Xen hypervisor under CloudStack orchestrator.

It is currently used with CloudStack ≥ 4.11 and XenServer 7 and build on top an Ubuntu Bionic base image.

Usage:

    bundle exec rake stemcell:build_with_local_os_image[cloudstack,xen,ubuntu,bionic,$PWD/tmp/ubuntu_base_image.tgz,"1.23"]

## Notes

1. This PR is derived from a series of patches we have been used for years in production on top the *Ubuntu Xenial* image which has reached it's end of life earlier this year; meaning it *does* have some Xenial leftovers in the code but is now build on top the Bionic releases.
2. While working on this PR for Bionic, we noticed the virtual machines network interface `eth0` was left in a "down" state when we used we used (`stemcell_builder/stages/bosh_go_agent/assets/bosh-agent-version`) a recent (`v2.366.0`) version of the bosh-agent; however it does work as expected with the `v2.346.0` version.

We would welcome any feedback on this pull request and additional requirements needed for it to be accepted.

Cheers,
Romain.

### DIFF
--- a/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
@@ -8,6 +8,19 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
   context 'installed by image_install_grub', {
     exclude_on_softlayer: true
   } do
+    context 'for cloudstack infrastructure and xen hypervisor', {
+        exclude_on_aws: true,
+        exclude_on_vcloud: true,
+        exclude_on_vsphere: true,
+        exclude_on_google: true,
+        exclude_on_warden: true,
+        exclude_on_azure: true,
+        exclude_on_openstack: true,
+    } do
+      describe file('/boot/grub/grub.cfg') do
+        its(:content) { should match ' console=hvc0' }
+      end
+    end
     describe file('/boot/grub/grub.cfg') do
       it { should be_file }
       its(:content) { should match 'set default="0"' }
@@ -35,6 +48,7 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
   context 'installed by image_install_grub_softlayer_two_partitions', {
       exclude_on_alicloud: true,
       exclude_on_aws: true,
+      exclude_on_cloudstack: true,
       exclude_on_google: true,
       exclude_on_vsphere: true,
       exclude_on_vcloud: true,
@@ -70,6 +84,7 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
     exclude_on_alicloud: true,
     exclude_on_aws: true,
     exclude_on_azure: true,
+    exclude_on_cloudstack: true,
     exclude_on_google: true,
     exclude_on_vcloud: true,
     exclude_on_vsphere: true,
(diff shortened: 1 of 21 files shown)


bosh-stemcell/spec/stemcells/ubuntu_xenial_spec.rb
```diff
@@ -13,7 +13,7 @@
       its(:content) { should match %r{kernel /boot/vmlinuz-\S+-generic ro root=UUID=} }
       its(:content) { should match ' selinux=0' }
       its(:content) { should match ' cgroup_enable=memory swapaccount=1' }
-      its(:content) { should match ' console=ttyS0,115200n8 console=tty0' }
+      its(:content) { should match %r{console=ttyS0,115200n8 console=(hvc|tty)0} }
```
ramonskie, line 16: its unnecessary to add the xenial related specs in the bionic track

ramonskie:
the issue related to the newer bosh agent can be found here: https://github.com/cloudfoundry/bosh-agent/pull/238

i also noticed that this has never been made availeble in xenial branch or even listed on bosh.io.
is this something that is required as well

romain-dartigues:
Thank you for your feedback, I will promptly take it into account (as well as the CLA issue).

bosh-stemcell/spec/stemcells/ubuntu_xenial_spec.rb
romain-dartigues: The PR has been adapted to remove Xenial specifics.

ramonskie:
we can accept this PR.
and if you want to have these stemcells builded and released on bosh.io
we probably need to discus this with the cloudfoundry foundation

bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
```diff
@@ -16,6 +16,7 @@
       its(:content) { should match ' selinux=0' }
       its(:content) { should match ' cgroup_enable=memory swapaccount=1' }
       its(:content) { should match ' console=ttyS0,115200n8' }
+      its(:content) { should match ' console=hvc0' }
```
rkoster, line 19: This change will probably break other infrastructure providers. Consider scoping it to CloudStack.

ramonskie: just ran a few tests and doesn't seem to break any tests although i would have expected it too break.

romain-dartigues: @rkoster < you are right, I did ran the tests and the builds only for our infrastructure and forgot that; let me fix it.

ramonskie: i tested it multiple times as i thought this would indeed brake other iaases. but no issues so far

romain-dartigues:
> we can accept this PR.
> and if you want to have these stemcells builded and released on bosh.io
> we probably need to discus this with the cloudfoundry foundation

Thank you for your support and insight on this PR.
If we can have it merged it will be plenty (as we can continue to build it locally until a larger number of users using the same kind of infrastructure make themselves known).

bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
rkoster: LGTM

PR #180, opened 2021-07-13
🐞 vCenter correctly displays IP address of VMs

cunnie:
This bug only affected VMs which run containers, for example Kubernetes worker nodes and Cloud Foundry Diego cells. The bug would manifest as an empty IP address on the vCenter UI in the VM's Summary page next to the "IP Addresses" field.

The cause appears to be the containers' network interfaces confusing `toolsd`, the executable of the "open-vm-tools" package which relays information about the VM to the vCenter.

This commit remedies this problem by directing `toolsd` to ignore container network interfaces (CNIs) owned by Silk and NSX-T Container Plug-in (NCP) by modifying its configuration file, `/etc/vmware/tools.conf`.

This is the list of NIC patterns we ignore. The first bullet is the [default pattern
list](https://github.com/vmware/open-vm-tools/blob/677ddf97a85f8b96b95ac60f54b59df264d5c4d5/open-vm-tools/tools.conf#L224-L225):

- `veth*,docker*,virbr*` (default)
- `silk-vtep,s-*` (Silk: Cloud Foundry container overlay network)
- `ovs*,erspan*,nsx-container,????????-????-*` (NCP: NSX-T Container Plug-in and Open vSwitch)

Note: `eth0` is typically the NIC which contains the correct IP address of the VM; all other NICs are suspect.

### DIFF
--- a/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
@@ -138,6 +138,11 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
     describe package('open-vm-tools') do
       it { should be_installed }
     end
+    describe file('/etc/vmware-tools/tools.conf') do
+      it { should be_file }
+      its(:content) { should match '\[guestinfo\]' }
+      its(:content) { should match 'exclude-nics=veth\*,docker\*,virbr\*,silk-vtep,s-\*,ovs\*,erspan\*,nsx-container,\?\?\?\?\?\?\?\?-\?\?\?\?-\*' }
+    end
   end
 
   context 'installed by system_softlayer_open_iscsi', {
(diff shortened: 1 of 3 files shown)


cunnie:
More background on this behavior: (not public viewable, sorry: https://pivotal-spike.atlassian.net/browse/PKS-693)

We've created test stemcells and deployed TAS with them, and checked whether we could view the IP addresses on the vCenter GUI:

- ✅ Xenial + Silk

klakin-pivotal:
So, I'm more than a little bit frustrated that we _have_ to do this (this smells like a... silly `open-vm-tools` bug to me), but the change looks sane, and you folks say that you've tested it, so into our testing pipelines it goes! (Thanks for the PR!)

There will be a bit of a delay in merging in the Bionic version of this PR... I don't remember exactly what we're _supposed_ to do with Bionic changes, so I'll need to ask around.

PR #181, opened 2021-07-13
🐞 vCenter correctly displays IP address of VMs

cunnie:
This bug only affected VMs which run containers, for example Kubernetes worker nodes and Cloud Foundry Diego cells. The bug would manifest as an empty IP address on the vCenter UI in the VM's Summary page next to the "IP Addresses" field.

The cause appears to be the containers' network interfaces confusing `toolsd`, the executable of the "open-vm-tools" package which relays information about the VM to the vCenter.

This commit remedies this problem by directing `toolsd` to ignore specific patterns of ethernet NICs when searching for the IP address of the VM; it does so by blacklisting NICs in the `/etc/vmware-tools/tools.conf` configuration file.

This is the list of NIC patterns we ignore. The first bullet is the [default pattern
list](https://github.com/vmware/open-vm-tools/blob/677ddf97a85f8b96b95ac60f54b59df264d5c4d5/open-vm-tools/tools.conf#L224-L225):

- `veth*,docker*,virbr*` (default)
- `silk-vtep,s-*` (Silk: Cloud Foundry container overlay network)
- `ovs*,erspan*,nsx-container,????????-????-*` (NCP: NSX-T Container Plug-in and Open vSwitch)

Note: `eth0` is typically the NIC which contains the correct IP address of the VM; all other NICs are suspect.

### DIFF
--- a/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_bionic_spec.rb
@@ -174,6 +174,11 @@ describe 'Ubuntu 18.04 stemcell image', stemcell_image: true do
     describe package('open-vm-tools') do
       it { should be_installed }
     end
+    describe file('/etc/vmware-tools/tools.conf') do
+      it { should be_file }
+      its(:content) { should match '\[guestinfo\]' }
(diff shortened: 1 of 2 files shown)


cunnie:
- related PR for Xenial (this is Bionic): <https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/180>

We've created test stemcells and deployed TAS with them, and checked whether we could view the IP addresses on the vCenter GUI:

- ✅ Bionic + Silk + vCenter 7

(We couldn't test Bionic + NCP + vCenter 7 because we didn't have that setup available to us)

PR #189, opened 2021-08-31
Add rsyslog-mmnormalize pkg, allow for better filtering

FloThinksPi:
We'd like to add rsyslog-mmnormalize package to allow for better filtering. As syslog-release of cf uses the OS rsyslogd the current way to go is to add it here.
This module is needed to properly filter the rather unconventional cloud_controller_ng syslogs format.
Usecase: We have is to just relay syslog messages for audit log purposes that contain altering http operations (PUT/DELETE etc.) and do not forward normal GET requests.


rkoster:
It seems like `rsyslog-mmnormalize` is not available for xenial, which if fine, but it would need some conditional logic around it. Also did a quick check on plain bionic docker image and got: `Unable to locate package rsyslog-mmnormalize`.

Also this link https://launchpad.net/ubuntu/bionic/+source/rsyslog/+changelog seems to suggest mmnormalize was dropped.

bgandon:
Indeed, on a plain Bionic Docker image, you need the `adiscon/v8-stable` Personal Package Archive (PPA):
```
apt install "software-properties-common"     # brings the 'add-apt-repository' utility
add-apt-repository "ppa:adiscon/v8-stable"   # install the Rsyslog v8-stable Personal Package Archive
```
This PPA is already present on Bionic Stemcells.

With version `8.2108.0` of Rsyslog, the `rsyslog-mmnormalize` package is still there. They also say so for Focal package: https://launchpad.net/ubuntu/focal/+source/rsyslog/+changelog

I suspect that by “drop” they actually mean “extract from main package to a distinct separate package” to reduce direct dependencies on the Rsyslog main package.

ramonskie:
its not necessary to specifically only set this the package `rsyslog-mmnormalize` for bionic
as this package is available for xenial,bionic,focal 
and this provided by the `ppa:adiscon/v8-stable`
which is added in the stage `base_ubuntu_packages`  
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/master/stemcell_builder/stages/base_ubuntu_packages/apply.sh#L40

therefor we can accept this PR.

the only problem i have now is that this is based on the master branch.
so this should be backported in the xenial and bionic branches

ramonskie (changes requested):
please add the package also in dpkg package list
`../bosh-stemcell/spec/assets/dpkg-list-ubuntu-xenial.txt`
`../bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt`

bgandon:
All right, so now the PR is properly rebased on `ubuntu-bionic/master` branch, and the Deb package list has been updated as requested.

When we backport this to Xenial, then we'll update the proper `dpkg-list-ubuntu-xenial.txt` file.

ramonskie:
thanks @FloThinksPi  @bgandon  for your PR

PR #197, opened 2021-10-29
621 nats agent cgroup iptables

nouseforaname:
Add IPTables rules to prevent processes other than the agent on a bosh deployed vm to connect to director nats.

The if-up.d script will parse out the the nats address from settings.json (once it is available). If the URL begins with 'nats://' (vm deployed by a running director) it adds the iptables rules to prevent access to the director nats. If the URL begins with 'https://' it does nothing (assuming it is a create env).

PR #198, opened 2021-10-29
456 nats agent cgroup iptables

nouseforaname:
Add IPTables rules to prevent processes other than the agent on a bosh deployed vm to connect to director nats.

The if-up.d script will parse out the the nats address from settings.json (once it is available). If the URL begins with 'nats://' (vm deployed by a running director) it adds the iptables rules to prevent access to the director nats. If the URL begins with 'https://' it does nothing (assuming it is a create env).


rkoster:
@nouseforaname could you apply the changes from the 621 PR to this PRs branch?

rkoster:
Changes have been reviewed and approved here https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/197

rkoster:
Thanks @nouseforaname

PR #204, opened 2021-11-19
Ubuntu bionic cleanup

toabctl:


toabctl:
I did a test build with `bundle exec rake --trace stemcell:build_os_image[ubuntu,bionic,/opt/bosh/tmp/ubuntu-bionic-pr-204.tgz]` and that passed.
Is there any CI involved for those kind of PRs ?

toabctl:
@rkoster maybe you could have a look given that you have a similar WIP branch for dropping `ixgbevf` custom installation?

rkoster:
For context:
- Original story from linked branch: https://www.pivotaltracker.com/n/projects/956238/stories/174635234
- Branch was never merged and instead the following commit was made against the bionic branch: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/fc0a38a419c3a57a664a8b6459330387bc6240c1

stemcell_builder/stages/base_apt/apply.sh
```diff
@@ -17,10 +17,6 @@ EOS
 # Upgrade systemd/upstart first, to prevent it from messing up our stubs and starting daemons anyway
 pkg_mgr install systemd
 
-# Normalize initramfs so that all distos can use dracut
-pkg_mgr purge busybox-initramfs
-pkg_mgr install dracut
```
rkoster, line 22: Why is dracut removed?

rkoster (changes requested):
It was no clear to me why dracut should be removed. Also from motivation for removing this package was not given in the PR description.

stemcell_builder/stages/base_apt/apply.sh
toabctl: @rkoster I updated the commit message. Quoting here the 2 reasons:
```
- initramfs tools is the default way on Ubuntu for creating initial
  ramdisk. That means that it's in the "main" component which is
  supported by the Ubuntu security team[1]. dracut is in universe
  which is not supported by the Ubuntu security team.
- the Ubuntu FIPS[0] packages can depend on initramfs tools. So
  installing those FIPS packages does not work with dracut.
```

toabctl: For context - I plan to propose more patches to be able to build a Ubuntu Bionic FIPS stemcell. I have a [POC](https://github.com/toabctl/bosh-linux-stemcell-builder/commit/7b9edb1b7b4478f95488102cd0e02754412c186a) for that and plan to introduce those changes with propper small commits. There are still some TODOs but the build & the stemcell in general works already.

ramonskie: i think we should keep FIPS enabled stemcells and the opensource ones separate.
we can use one repo but then we can use an extra stage.
where it will configure everything FIPS releated

rkoster: I agree with @ramonskie, since FIPS support requires a `Ubuntu Advantage` subscription we can not make the OSS stemcells FIPS complaint by default. So these stemcells would need to be built manually, making this easier by having a flag or target in the stemcell builder would definitely be an option here.

With regards the removal of `dracut`, it is unclear to me at this point if other releases might depend on this. So I'm considering this a breaking change for now. I don't see a lot of value in making this change in bionic but it would be a great time to submit this change for jammy now (since that's still in alpha, the branch can be found [here](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/tree/ubuntu-jammy/master)).

The rest of the PR looks good to me, so if you could remove the changes related to switching from dracut to initframfs, this PR can be merged.

PS: Also found that at some point in time we switched from initframfs to dracut: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/eb7c3c441d1c05476c2fd7b6a2eb1c69eeae8ad2 unfortunately there are no details provided on their motivation.

toabctl: Yes, the idea is to have most of the FIPS related code in an exta stage (see https://github.com/toabctl/bosh-linux-stemcell-builder/tree/7b9edb1b7b4478f95488102cd0e02754412c186a/stemcell_builder/stages/base_apt_fips ).
But having **all** FIPS related code in a single stage is difficult. In some stages, there would need to be an `if THIS_IS_FIPS; then do X; else do Y;`
Would that kind of `if/else` be ok in other stages?
Eg. this `dracut` installation is one of theses stages that would not work for FIPS (because this would remove all the FIPS packages again)
Another option would be to have the `if/else` in the stage_collection and have (where needed) `default` and `fips` stages.

toabctl: @rkoster split out the change for dropping `ixgbevf` manual installation into https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/206 (bionic) and https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/207 (jammy).

toabctl: I'll PR the dracut removal for Jammy when #207 got merged.

ramonskie (approved):

rkoster:
@toabctl https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/207 has been merged

rkoster (approved):

rkoster:
Thanks! @toabctl

toabctl:
@rkoster @ramonskie I don't want to complain, but I thought you **don't** want to merge **this** PR? Now dracut support got removed for bionic.
Don't get me wrong - I'm happy that it got removed. Just want to make sure that this is really what you wanted to do?!

p.s. https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/208 is the dracut removal for `ubuntu-jammy/master` .

stemcell_builder/stages/base_apt/apply.sh
toabctl: dracut removal for Jammy is now https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/208

rkoster:
@toabctl yeah we talked it over a bit more and decided it's okey for bionic.

PR #208, opened 2021-12-02
Drop dracut and switch to default initramfs tools

toabctl:
From what I can see, dracut was used to normalize initramfs creation
for all distros. But it seems since commit 6acd0c9cf71ed that only
Ubuntu is left.
So switch to the default initramfs tools for creating the initial
ramdisk.
This has multiple advantages:
- initramfs tools is the default way on Ubuntu for creating initial
  ramdisk. That means that it's in the "main" component which is
  supported by the Ubuntu security team[1]. dracut is in universe
  which is not supported by the Ubuntu security team.
- the Ubuntu FIPS[0] packages can depend on initramfs tools. So
  installing those FIPS packages does not work with dracut.

[0] https://ubuntu.com/security/certifications/docs
[1] https://wiki.ubuntu.com/SecurityTeam/FAQ

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
@@ -52,8 +52,6 @@ dmsetup
 dnsutils
 dpkg
 dpkg-dev
-dracut
-dracut-core
 e2fsprogs
 eject
 fdisk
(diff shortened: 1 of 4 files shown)


rkoster:
Change was already approved here: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/204

rkoster (approved):

rkoster:
Thanks! @toabctl

PR #214, opened 2022-01-10
Install netcat-openbsd explicitly

toabctl:
netcat-openbsd is needed (see content of
bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt) so install it
explicitly.
This fixes a problem when using Ubuntu FIPS packages where
netcat-openbsd is not automatically installed.


toabctl:
Looks like it's not needed. the package is included.

PR #210, opened 2021-12-10
Add stage to build a Ubuntu Bionic FIPS stemcell

toabctl:
There are usecases where FIPS compliant stemcells are needed. So
adding an option to be able to build FIPS compliant stemcells here.

Note: The implementation depends on the Operating system variant and the UBUNTU_ADVANTAGE_TOKEN
environment variable. If the variant is `fips` and the env variable is set (it needs to be a valid
token because it's later used by the ubuntu-advantage tool to install
the relevant FIPS packages), the build will produce a FIPS stemcell.

PR #215, opened 2022-01-13
Add support for building stemcell variants

rkoster:
This PR is an providing a framwork to be used by the fips PR: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/210.

To build a variant suffix the version with -{variant_name} example:
```
bundle exec rake stemcell:build_os_image[ubuntu,bionic-fips,$PWD/tmp/ubuntu_base_image.tg]
```

The variant will also be passed to rspec as a tag which can be used to run specific tests only when building a variant.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
@@ -183,7 +183,7 @@ module Bosh::Stemcell
       else
         nil
       end,
-      unless operating_system.variant.nil?
+      if operating_system.variant?
         " --tag #{operating_system.variant}"
       end,
       ].compact.join(' ').rstrip
--- a/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
@@ -181,9 +181,12 @@ module Bosh::Stemcell
       when 'google'
         ' --tag ~exclude_on_google'
       else
-        ''
+        nil
       end,
-      ].join(' ').rstrip
+      unless operating_system.variant.nil?
+        " --tag #{operating_system.variant}"
+      end,
+      ].compact.join(' ').rstrip
     end
 
     def image_file_path
--- a/bosh-stemcell/lib/bosh/stemcell/operating_system.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/operating_system.rb
@@ -10,6 +10,7 @@ module Bosh::Stemcell
 
     class Base
       attr_reader :name, :version, :variant
+      alias_method :variant?, :variant
 
       def initialize(options = {})
         @name = options.fetch(:name)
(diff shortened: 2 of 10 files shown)


bosh-stemcell/lib/bosh/stemcell/build_environment.rb
```diff
@@ -181,9 +181,12 @@ def exclude_exclusions
       when 'google'
         ' --tag ~exclude_on_google'
       else
-        ''
+        nil
       end,
-      ].join(' ').rstrip
+      unless operating_system.variant.nil?
```
jpalermo: My small caveman brain has trouble parsing this. Is this the same as `if operating_system.variant`?

Functionally it is different because `false` values would behave differently in the two, but that's probably not a reasonably valid value for `variant` I'm guessing

bosh-stemcell/lib/bosh/stemcell/operating_system.rb
```diff
@@ -23,7 +24,8 @@ def ==(other)
 
     class Ubuntu < Base
       def initialize(version)
-        super(name: 'ubuntu', version: version)
+        (version, variant) = version.split('-') unless version.nil?
```
jpalermo: Another double negative. This one my brain has an easier time parsing, but `if version` still might be better?

bosh-stemcell/lib/bosh/stemcell/build_environment.rb
```diff
@@ -181,9 +181,12 @@ def exclude_exclusions
...
+        " --tag #{operating_system.variant}"
+      end,
+      ].compact.join(' ').rstrip
```
jpalermo, line 189: The `exclusions` for specs to run has also confused me a number of times, so I'm probably not fully understanding what's going on here. But are we creating an exclusion here, or an inclusion? All the other tags use a `~` to exclude certain specs, but here we're just doing `--tag variant` which will exclude everything except specs tagged with the variant?

I suppose that would be excluding everything EXCEPT the variant? So maybe still technically an "exclusion" of sorts.

bosh-stemcell/lib/bosh/stemcell/build_environment.rb
rkoster: Yeah, I guess you are right, getting rid of some of the explicitness in order to improve readability would make the code more idiomatic.

bosh-stemcell/lib/bosh/stemcell/build_environment.rb
rkoster: Was hoping this would allow running additional tests only when building a `fips` stemcell. But it does not seem to work like that according to the docs: https://relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option. Will make it an exclusion filter so that at least when building a fips stemcell certain tests could be excluded when needed.

jpalermo (approved):

ramonskie (approved):

PR #203, opened 2021-11-17
Nats agent cgroup iptables  cherry-picks for bionic branch

nouseforaname:

PR #216, opened 2022-01-25
Reintroducing Nats iptables firewall 

nouseforaname:
The original [PR](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/203) contained bugs and was reverted. These bugs have now been fixed by @cunnie, @ragaskar and @ystros

🐞 NATS isolation works properly on GCP et al.
    - Script is inline instead of defining a function then calling said
      function.
    - No backgrounding.
    - Logging works properly, and appends multiple runs (`&2>` → `&>>`)
    - Log more stuff (`set -x`), but exclude passwords.
    - Rather than wait forever for `/var/vcap/bosh/settings.json`, we time
      out after 15 seconds
    - More descriptive variable names (`BOSH_PORT` → `DIRECTOR_NATS_PORT`)
    - Updates parsing of NATS URL to handle cases where there's no
      `user:password@` embedded.

    [#180698114](https://www.pivotaltracker.com/story/show/180698114)

    Signed-off-by: Brian Cunnie <bcunnie@vmware.com>


stemcell_builder/stages/bosh_go_agent/assets/02-restrict-nats-api-access-allow-monit-api
```diff
@@ -0,0 +1,78 @@
...
+  echo "NATS Address couldn't be parsed into IP:PORT, this script is failing to interpret the /var/vcap/bosh/settings.json"
+  exit 1
+fi
+
+# GET FIRST FIELD WITH DELIMITER `:` LEAVING `IP`
+DIRECTOR_IP=$( cut -f1 -d: <<< "${DIRECTOR_NATS}" )
+# GET SECOND FIELD WITH DELIMITER `:` LEAVING `PORT`
+DIRECTOR_NATS_PORT=$( cut -f2 -d: <<< "${DIRECTOR_NATS}" )
+
+echo "$(date) restricting director NATS access to bosh-agent"
```
beyhan, line 66: This won't work so well when `director_nats` is configured with a hostname. In this situation following can happen:

1. `iptable` rule is executed and resolves the `hostname` to an `ip` and uses it for the rule
2. directore is recreated with another `ip` and hostname gets updated
3. the `iptable` rule still blocks the access to the old `ip` and every process on the VM can access nats under the new `ip` until the VM is recreated or agent gets restarted
4. when agent is restarted, the agent will trigger the network setup again (if I'm not mistaken here) and a new `iptable` rule will be set with the new bosh `ip`.

What is your comment on this?

jpalermo:
Just FYI on this, we're still seeing this fail to successfully block nats traffic on our Xenial stemcells. So it's not ready to be merged yet.

stemcell_builder/stages/bosh_go_agent/assets/02-restrict-nats-api-access-allow-monit-api
jpalermo: My opinion is this is still a much smaller attack surface area than we've had for the last 10 years.

We could possibly do a follow up to try and address this in some way, although we're having trouble making the current functionality work, so adding additional complexity does make me a bit nervous.

beyhan: I also agree that we shouldn't add more complexity here. For the described situation a recreate of all VMs could be recommended e.g.

PR #222, opened 2022-02-04
Revert "Reintroducing Nats iptables firewall Ubuntu jammy/master"

nouseforaname:
Reverts cloudfoundry/bosh-linux-stemcell-builder#219

We decided to move this feature to the agent since the stemcell implementation is not reliable


ramonskie:
this was already reverted

rkoster:
Closing because already reverted here: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/88297dca30c848488493a141880f6a561d2237e5

PR #228, opened 2022-04-26
NATS access helper aids troubleshooting

cunnie:
Newest stemcells prevent processes from accessing NATS (port 4222) on the BOSH Director because security; however, this can hamper troubleshooting.

This commit adds a helper script to enable an operator to access the NATS port on the director; it accomplishes this by adding the PID to the cgroup which is allowed to communicate with the Director's NATS server.


ramonskie (approved):

ramonskie (changes requested):
.

rkoster:
@nouseforaname could you update this PR with your concerns?

ragaskar:
We think the documentation on bosh.io regarding getting access to NATS is simpler to use/maintain over having this script, so we're going to simply close out this PR.

Issue #231, opened 2022-06-30
Azure - SSH host key generation failed


MSSedusch:
In some cases, Azure VM deployments fail with the following error:

`
Error message: {
"startTime": "2022-06-24T02:11:16.5239158+00:00",
"endTime": "2022-06-24T02:13:05.4620971+00:00",
"status": "Failed",
"error": {
"code": "OSProvisioningClientError",
"message": "SSH host key generation failed. Error details: [ProvisionError] Failed to generate ssh host key: ret=255, out= ssh-keygen: /etc/ssh/ssh_host_rsa_key.pub: No such file or directory\n. To resolve this issue verify if Linux agent is set up properly. \r\n    You can check the instructions at : https://azure.microsoft.com/documentation/articles/virtual-machines-linux-agent-user-guide/\r\n   "
}
`

The root cause is a race condition between rc.local and the walinuxagent. rc.local re-creates the host SSH keys and walinuxagent tries to use them. If they are not created yet, the VM deployment fails. I will open a PR to fix it.

ramonskie:
#232 #233 are merged
we can close this issue

PR #232, opened 2022-06-30
add dependency on first boot task (rc.local)

MSSedusch:
Fix for #231


jpalermo:
Thanks for the PR @MSSedusch, is this likely something we should also merge into the jammy branch?

MSSedusch:
I am testing it right now with jammy. Seems like there is more work to do for jammy. Can we already merge this PR for bionic?

jpalermo:
Yeah, sorry for the delay on this, I haven't had a chance to look at PRs recently.

jpalermo (commented):
👍 Thanks!

Issue #234, opened 2022-07-15
Cannot create Azure VM with Ubuntu Jammy stemcell


MSSedusch:
While working on #231 I also tested this fix for Ubuntu Jammy but provisioning of a VM would always time out.
I also tested the latest Ubuntu Jammy stemcell from the official repository (https://bosh.cloudfoundry.org/stemcells/) with the same result.

The root cause seems to be that the network does not work after the first boot of the VM. I tested with Accelerated Networking enabled and disabled.

If I connect to the VM via the serial console and execute ifdown & ifup, the VM can connect to the internet and the Azure wireserver (to report health).

Has anyone tested the Ubuntu Jammy stemcell on Azure?

ystros:
We've been seeing similar behavior. Tracing through the Window Azure Linux Agent source code yesterday, we noticed that it is calling `ifdown eth0 && ifup eth0` as part of its provisioning code. It ends up marking the VM as provisioned, but our theory is that the network never comes back up and it gets stuck trying to report back to the Azure API.

See:
* Provisioning method: https://github.com/Azure/WALinuxAgent/blob/v2.6.0.2/azurelinuxagent/pa/provision/default.py#L213
* `publish_hostname`: https://github.com/Azure/WALinuxAgent/blob/v2.6.0.2/azurelinuxagent/common/osutil/default.py#L1176-L1180
* `restart_if`: https://github.com/Azure/WALinuxAgent/blob/v2.6.0.2/azurelinuxagent/common/osutil/default.py#L1163-L1174

Relevant logs from the VM. You can see the network go down then supposedly come up. Also of note (may not be relevant), some of the messages from waagent come after it runs `ifup eth0`, but the corresponding `systemd-networkd[302]: eth0: Link UP` comes after the last waagent message. waagent never makes it to the "Provisioning complete" message, nor a "Provisioning failed" message.
```
Jul 14 18:48:37 bosh-stemcell python3[1686]: 2022-07-14T18:48:37.879806Z INFO Daemon Daemon Azure Linux Agent Version:2.6.0.2
Jul 14 18:48:37 bosh-stemcell python3[1686]: 2022-07-14T18:48:37.884777Z INFO Daemon Daemon OS: ubuntu 22.04
Jul 14 18:48:37 bosh-stemcell python3[1686]: 2022-07-14T18:48:37.888403Z INFO Daemon Daemon Python: 3.10.4
[...]

PR #235, opened 2022-07-19
switch to cloud-init for vm provisioning

nouseforaname:
on jammy waagent interferes with our own provisioning in bosh-agent.
this switches off waagent as a provisioning tool but keeps the
extensions feature of waagent. it uses a bare minimum cloud-init setup
to pull out metadata and update networking of the vm.

its explicitly switches off modules that would handle disks and other
cloud-init features that would interfere with bosh agent jobs.

it updates the tests to consider additional packages that are getting
installed for cloud init

it also changes the agent.json to use the paths where cloud-init will
extract the userdata and removes registry usage (not necessary and also
deprecated) while at it.

Even Microsoft recommends using cloud-init.

https://docs.microsoft.com/en-us/azure/virtual-machines/linux/create-upload-ubuntu


stemcell_builder/stages/system_azure_wala/apply.sh
```diff
@@ -46,3 +45,24 @@ cat > $chroot/etc/logrotate.d/waagent <<EOS
     missingok
 }
 EOS
+
+#setup cloud-init
+rm $chroot/etc/cloud/*.cfg
+rm $chroot/etc/cloud/cloud.cfg.d/*.cfg
+cp -f $dir/assets/etc/cloud-init.cfg $chroot/etc/cloud/cloud.cfg
+cp -f $dir/assets/etc/*-*.cfg $chroot/etc/cloud/cloud.cfg.d/
+
```
jpalermo, line 54: Couple of things, (we started to do them ourselves, but then got a bit confused).

Could we rename `system_azure_wala` to something like `system_azure_init`, since it's doing both the walinux agent and cloud-init now? Makes it less surprising when you find cloud-init things going on in the stage.

It might also help clarity to separate the config files in `assets/etc`, maybe add a `waagent` and `cloud-init` subdirectories.

We noticed that due to the `*-*.cfg` the `90_datasource.cfg` file isn't making it onto the stemcell. We're not sure what this file is doing, but it seemed weird that everything would work without it.

`cloud-init.cfg` also gets copied both into `/etc/cloud` and `/etc/cloud/cloud.cfg.d/`. We're guessing it's only supposed to be in one of the two spots.

stemcell_builder/stages/system_azure_wala/assets/etc/cloud-init.cfg
```diff
@@ -0,0 +1,57 @@
+users:
+   - default
```
jpalermo: Seems like an extra indent snuck in here

jpalermo (commented):
We also squashed the two commits into a single one and expanded the commit message to add some more details to what's going on.

stemcell_builder/stages/bosh_azure_agent_settings/apply.sh
```diff
@@ -16,24 +16,14 @@ cat > $chroot/var/vcap/bosh/agent.json <<JSON
...
       "Sources": [
-        {
-          "Type": "ConfigDrive",
-          "DiskPaths": [
-            "/dev/disk/by-label/azure_cfg_dsk",
-            "/dev/disk/by-label/AZURE_CFG_DSK"
-          ],
-          "MetaDataPath": "configs/MetaData",
-          "UserDataPath": "configs/UserData"
-        },
```
jpalermo, line 27: There does still seem to be code in the azure cpi for [ConfigDrive](https://github.com/cloudfoundry/bosh-azure-cpi-release/blob/master/src/bosh_azure_cpi/lib/cloud/azure/disk/config_disk_manager.rb), but from what we can tell, it was added by Microsoft back in 2018, then 2 months later the `spec` property was removed, so you've never really been able to use a ConfigDrive. We're not really sure why it was added, but it seems to never be used, so this removal should be safe (and Jammy seems like a good chance to rip it out)

stemcell_builder/stages/system_azure_wala/apply.sh
nouseforaname: nice catch about the datasource. I guess it was saved by some smarts // default in cloud-init. This should've explicitly set the datasource for user/metadata to azure (same as bosh agent cloud-init will source the info via different codepaths). setting it explicitly to azure should at least (probably minimally) decrease detection time and make it more reliable.

nouseforaname:
solves #234 


Ran bats for azure with unmanaged and managed disks, both passed

stemcell_builder/stages/system_azure_init/apply.sh
```diff
@@ -5,7 +5,7 @@ set -e
 base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
 
-packages="python3 python3-pyasn1 python3-setuptools"
+packages="python3 python3-pyasn1 python3-setuptools python-is-python3 cloud-init"
```
ramonskie, line 8: discussed about the `python-is-python3` package
and if it should be installed globally on all stemcells as python3 is installed by default on all stemcells.
but as azure is the only one using python in our stemcells anyways it seems to make sense to put it here

PR #236, opened 2022-07-22
Enable FIPS mode for bionic-fips stemcells

toabctl:
To have FIPS really enabled (/proc/sys/crypto/fips_enabled needs to
be 1), the kernel command line needs to have a fips=1
parameter. That's now set.


ramonskie (approved):
lgtm

PR #243, opened 2022-10-12
Permissions fix  for `ubuntu-advantage` cache file

aramprice:
The change in this commit updates the `esm_enable` function so that it modifies the file permssions of
`/tmp/ubuntu-advantage/candidate-version` created by `ubuntu-advantage` from `777` -> `664` by stripping all execute permissions `a-x`, and removing write permissions for "other" `o-w`.

This is done so that the "stig: V-38643" test which validates that there are no "other" writeable files on the filesystem.

Once the file has been created the `ubuntu-advantage` tool does not appear to modify the permissions however the `esm_enable` function (re)sets the permissions on every invocation.

Context: the `ubuntu-advantage` / `ua` utility appears to be writing a file with `777` permissions, most likely because of [this commit](https://github.com/canonical/ubuntu-advantage-client/commit/ce808efa9fe4120e81d0424f4d9978da2f00cb78).

### DIFF
--- a/stemcell_builder/lib/prelude_apply.bash
+++ b/stemcell_builder/lib/prelude_apply.bash
@@ -112,6 +112,7 @@ function esm_enable {
   pkg_mgr install ubuntu-advantage-tools
   run_in_chroot $chroot "ua attach ${ESM_TOKEN} --no-auto-enable"
   run_in_chroot $chroot "ua enable esm-infra"
+  run_in_chroot $chroot "chmod a-x,o-w /tmp/ubuntu-advantage/candidate-version"
 }
 
 function esm_disable {


stemcell_builder/lib/prelude_apply.bash
```diff
@@ -112,6 +112,7 @@ function esm_enable {
   pkg_mgr install ubuntu-advantage-tools
   run_in_chroot $chroot "ua attach ${ESM_TOKEN} --no-auto-enable"
   run_in_chroot $chroot "ua enable esm-infra"
+  run_in_chroot $chroot "chmod a-x,o-w /tmp/ubuntu-advantage/candidate-version"
```
klakin-pivotal, line 115: Yeah, I'd replace this with

```
run_in_chroot $chroot "rm -rf /tmp/ubuntu-advantage"
```
I find it hard to believe that any Ubuntu Advantage software would require files in `/tmp` after line 114.

aramprice: Out of an abundance of caution I'd prefer to stick with `chmod` rather that removal. This is mostly because I don't have a lot of confidence that the tool won't evolve to expect more things in this directory.

I think you are right about the file likely not being needed, especially given that the constant holding the path has ["cache" in the name](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/version.py#L51-L69).

This functionality appears to have been first added in [this commit](https://github.com/canonical/ubuntu-advantage-client/commit/ce808efa9fe4120e81d0424f4d9978da2f00cb78)

I also raised an issue on about the loose permissions on the file https://github.com/canonical/ubuntu-advantage-client/issues/2294

aramprice:
I'm going to merge this more cautious approach. If we get more info, or a resolution from upstream we can adjust our approach.

PR #247, opened 2022-10-27
Xenial - open-vm-tools

danielfor:
Canonical did not provide a patch for a recent [CVE](https://ubuntu.com/security/notices/USN-5578-1) in open-vm-tools.
Because of this we don't want to use their package, but compile this software with the latest release from vmware.
This PR is to make that change.


cunnie (approved):
Nice commit messages; glad you remove the dev files after they're no longer needed.

stemcell_builder/stages/system_open_vm_tools/apply.sh
```diff
@@ -6,14 +6,59 @@ base_dir=$(readlink -nf $(dirname $0)/../..)
...
+  cp $dir/assets/lib/systemd/system/vgauth.service $chroot/lib/systemd/system/vgauth.service
+  cp $dir/assets/lib/systemd/system/open-vm-tools.service $chroot/lib/systemd/system/open-vm-tools.service
+  mkdir $chroot/etc/systemd/system/open-vm-tools.service.requires/
+  run_in_chroot $chroot "
+    ln -s /lib/systemd/system/open-vm-tools.service /etc/systemd/system/multi-user.target.wants/open-vm-tools.service
+    ln -s /lib/systemd/system/vgauth.service /etc/systemd/system/open-vm-tools.service.requires/vgauth.service
+    "
+}
 
+if [[ "${DISTRIB_CODENAME}" == "xenial" ]]; then
```
aramprice, line 51: Since this code lives on a `xenial` branch would it make more sense to test something like "distro name", e.g. `ubuntu`?

aramprice: Looks like we switch on Ubuntu code-name all over the codebase so probably fine since `xenial` implies `ubuntu` even though it is confusing to me at first glance.

stemcell_builder/stages/system_open_vm_tools/apply.sh
```diff
@@ -6,14 +6,59 @@ base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
 source $base_dir/lib/prelude_bosh.bash
 
-# Installation on CentOS requires v7
-pkg_mgr install open-vm-tools
+function compile_package {
+  run_in_chroot $chroot "
+    cd /tmp
+    apt install -y autoconf \
```
aramprice, line 12: Consider using `apt-get` instead of `apt`. The `apt` command line utility regularly warns that its interface is not stable and that it should not be relied on for scripting.

We don't appear to care about the output of the `apt` command in this script, though if we did in the future it could be an issue.

aramprice (approved):

stemcell_builder/stages/system_open_vm_tools/apply.sh
danielfor: Thanks for the review, I was confused by that because we have branches for each OS, and also branches in the code.

stemcell_builder/stages/system_open_vm_tools/apply.sh
aramprice: https://askubuntu.com/questions/445384/what-is-the-difference-between-apt-and-apt-get
> The `apt` command is meant to be pleasant for end users and does not need to be backward compatible like apt-get(8).

PR #248, opened 2022-10-27
Xenial - open-vm-tools 456.x

danielfor:
Canonical did not provide a patch for a recent [CVE](https://ubuntu.com/security/notices/USN-5578-1) in open-vm-tools.
Because of this we don't want to use their package, but compile this software with the latest release from vmware.
This PR is to make that change.


cunnie (approved):
Nice commit messages, nice cleanup after install.

PR #251, opened 2022-11-01
Jammy stemcells include LLVM Clang Compiler

cunnie:
Ruby programs using Ruby compiled with GCC (GNU Compiler Collection) on Jammy stemcells have a much larger RSS (Resident Set Size) memory footprint, which can cause memory pressure. This affects Ruby-based programs such as the BOSH Director and the BOSH Azure, AWS, and vSphere CPIs. This can cause BOSH operations such as “bosh deploy” to take much longer and even time out.

This commit addresses the issue by including the GCC alternative, Clang, on the stemcell. When Ruby is compiled with the Clang compiler, it has a smaller memory footprint, and also has performance enhancements as well: `bosh deploy` can be up to twice as fast than on a Xenial-based GCC-compiled Director.

We include the native Ubuntu-packaged Clang on the stemcell rather than creating a Clang BOSH package because the compile times of the Clang BOSH package were incredibly long: We estimate a compile time of 150 minutes (2½ hours!) for a 2 vCPU 4 GiB BOSH Director using `bosh create-env`, over ten times longer than the next-longest package's compile time.

We install Clang as part of the "build-essential" step. "build-essential" is an Ubuntu package that includes the basic development tools, including the C compiler, GCC. We felt that Clang, a GCC-compatible compiler, should be installed in the same step.

The addition of the Clang compiler consumes ~400 MiB on the root filesystem; however, this space can be freed up by setting `env.bosh.remove_dev_tools` to `true`.

Drivebys:

- The README had the wrong location for the ovftool; now it has the correct location


cunnie:
**Don't merge yet**: It doesn't remove Clang when `remove_dev_tools` is set to true.

Issue #249, opened 2022-10-27
rsyslog crash on bosh-google-kvm-ubuntu-jammy-go_agent/1.18


nicregez:
### Summary

I suggest to consider and evaluate rsyslog-openssl as replacement for rsyslog-gnutls.

### Background

We had an `rsyslog` crash on one of our bosh managed VMs recently. This was not the first time the `rsyslog` daemon crashed, but unfortunately, I do not have more details on these older incidents. 

```
# journalctl | grep "Oct 23 05:00:41" | head -1
Oct 23 05:00:41 <vm guid> rsyslogd[263401]: omrelp[34.122.98.95:10515]: error 'TLS record reception failed [gnutls error -54: Error in the pull function.]', object  'conn to srvr 34.122.98.95:10515' - action may not work as intended [v8.2112.0 try https://www.rsyslog.com/e/2353 ]
```

I looked up [OMRELP Manual](https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html) which says that
> If you have problems with the default setting, we recommend to switch to “openssl”.

This is the [Reference for Error 2353](https://www.rsyslog.com/?s=error+2353).

rkoster:
@Benjamintf1 Would you have a problem with the Jammy stemcell switching to openssl as a tls backend for rsyslog?

rkoster:
cc @ctlong

Benjamintf1:
The most immediate and quick answer is...I don't know?

The longer version is, off the top of my head and through immediate research, I don't have an inclination to say which is better or worse. I think I experimented with it and had trouble when trying to fix jammy, but I fixed it by fixing other settings on the receiver which seemed to have changed. 

The easiest fix would be to setup a option to switch to openssl as the backend, but I'm not sure if that's sufficient to solve the problem or if it would require larger fixes. We have a pipeline setup with tests running against xenial, bionic, and jammy, so we should be able to run tests against any pr to insure it works on all of those stemcells at the moment. 

Broadly speaking, I'm pretty amiable to different solutions, as long as it passes the existing ci.

Benjamintf1:
We looked into it a little bit deeper, not only would you likely have to change quite a bit of the configuration to match the new network driver in syslog-release, but our current way of using the `obsolete legacy directives` in syslog-release are not even documented to work in the omrelp documentation. 

[...]

PR #254, opened 2022-11-03
fix rsyslog crash in case of connection abort

h0nIg:
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/249

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-bionic.txt
@@ -404,6 +404,7 @@ rng-tools
 rsync
 rsyslog
 rsyslog-gnutls
+rsyslog-openssl
 rsyslog-mmjsonparse
 rsyslog-mmnormalize
 rsyslog-relp
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
@@ -441,6 +441,7 @@ rpcsvc-proto
 rsync
 rsyslog
 rsyslog-gnutls
+rsyslog-openssl
 rsyslog-mmjsonparse
 rsyslog-mmnormalize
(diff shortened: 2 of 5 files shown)


ramonskie:
it seems that $ModLoad is the legacy use of loading a module.
maby we need to revise this whole config file. can yo confirm this?
i can fix this in a later stage but just wanted some confirmation

h0nIg:
@ramonskie adjustment is not required, even ubuntu did not adjust the configuration. On top of that rainerscript and the old syntax will co-exist: https://rainer.gerhards.net/2008/02/introducing-rainerscript-and-some-rsyslog-futures.html

> I hope you find the choice of name acceptable — and the overall direction we are taking a good one. As a side-note please let me assure that rsyslog’s easy and simple sysklogd-like configuration will not go away. Rainerscript will fully support that syntax, as well as all currently existing $-configuration lines.

Just checked the current rsyslog.conf and it uses both old/new syntax:

```
......
#################
#### MODULES ####
#################

module(load="imuxsock") # provides support for local system logging
#module(load="immark")  # provides --MARK-- message capability

# provides UDP syslog reception
#module(load="imudp")
#input(type="imudp" port="514")

# provides TCP syslog reception
#module(load="imtcp")
#input(type="imtcp" port="514")

# provides kernel logging support and enable non-kernel klog messages
module(load="imklog" permitnonkernelfacility="on")

###########################
#### GLOBAL DIRECTIVES ####
###########################

#
# Use traditional timestamp format.
# To enable high precision timestamps, comment out the following line.
#
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
......
```

ramonskie (approved):
LGTM

ramonskie:
does this also not need to be merged in to bionic?

h0nIg:
@ramonskie https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/255

rkoster (approved):

rkoster:
@h0nIg Thanks!

PR #255, opened 2022-11-07
fix rsyslog crash in case of connection abort

h0nIg:


ramonskie:
can be merged once #254 is reviewed

PR #257, opened 2022-11-09
update xenserver guest tools for jammy stemcell

psycofdj:
This PR implements jammy compatibility for stemcells using xenserver hypervisor.

Since the package `xe-guest-utilities` is no longer available in apt for jammy steamline, Xenserver guest utilities should now be compiled from source. 

The PR also includes a copy the Citrix license as required when redistributing xen-guest-utilities binaries.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -296,6 +296,7 @@ SECTION 1: BSD-STYLE, MIT-STYLE, OR SIMILAR STYLE LICENSES
    >>> webmock-1.11.0
    >>> yajl-ruby-0.8.2
    >>> yajl-ruby-1.1.0
+   >>> xe-guest-utilities-7.30.0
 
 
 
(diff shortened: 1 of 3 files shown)


ramonskie:
as we don't have the option to actually test this as we don't have a cloudstack account for testing.
i trust you that you tested this.

the only thing i don't like is the specific commit for `golang.org/x/sys`
is it possible to correct this to an actual version? `v0.2.0` seems to be the latest now

psycofdj:
Thanks. 

Indeed, this has been tested on our staging environments made of bosh, cf, concourse and prometheus deployments with cloudstack `4.16.x TLS`.  The updated version of `xe-guest-utilities` works as intended on Citrix XenServer and its opensource version XCP-ng.

For `golang.org/x/sys`, I ended up reworking the compilation directory tree. `xe-guest-utilities` has a very non-standard makefile based compilation process which is kinda hard to understand. In the end, I only need to vendor dependencies, the `golang.org/x/sys` version specified in `go.mod` is used. 

I'm marking the PR as WIP until I can validate that everything builds fine.

psycofdj:
everything is ok !

ramonskie (approved):
LGTM

rkoster:
Thanks! @psycofdj

PR #263, opened 2022-11-24
Installing ethtool in Jammy

danielfor:
Now the stemcell ships with ethtool, which is a tool that is occasionally useful to users facing networking problems.

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
@@ -63,6 +63,7 @@ dwz
 e2fsprogs
 eject
 emacsen-common
+ethtool
 fdisk
 file
 findutils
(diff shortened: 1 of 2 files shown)


ramonskie:
should we not keep the stemcell as lean as possible without introducing tools for all sort of could be issues?
as its rather easy to install it on a need to basis.
i would even suggest a bosh addon like https://github.com/cloudfoundry-community/toolbelt-boshrelease
for debugging purposes

danielfor:
Hi @ramonskie , thanks for your insights. Yes, we should keep them as lean as possible. We thought it might be better on this case to include `ethtool` in the stemcell because some users have restrictive environments in which the VMs cannot access the internet, so they can't install software with ease. The use case for this tool is to increase the ringbuffer size on some big VMs (Diego cells) and the current toolbelt does not have tools for that. Having ethtool might make the modification easier for them.

klakin-pivotal:
> ...without introducing tools for all sort of could be issues?

@ramonskie These users and issues are not theoretical. Our VMware support folks have run into multiple cases where having `ethtool` would have been very helpful, but was not readily available because the affected VMs were running in an environment that did not and could not have Internet access.

cunnie:
@ramonskie I wholeheartedly agree: "lean as possible"

I know that on at least two occasions in my previous life as a sysadmin I had to use ethtool, so I think it's worth the 672kB.

rkoster (approved):

Issue #239, opened 2022-09-20
vim defaults broken on jammy


domdom82:
While testing jammy I've noticed this bug:
```
# vim
E1187: Failed to source defaults.vim
Press ENTER or type command to continue
```

There is an upstream discussion about the issue here:

https://www.mail-archive.com/ubuntu-bugs@lists.ubuntu.com/msg6018994.html

I've opened this issue to track / possibly fix the upstream issue as it severely hampers productivity with jammy stemcells.

ramonskie:
the bug is confirmed at canonical see https://bugs.launchpad.net/ubuntu/+source/vim/+bug/1968912
a workaround would be to add the `vim-runtime` package

as vim-tiny is installed by default and when installing `vim-runtime` its not tiny vim anymore ;)

or we wait until canonical fixed there vim-common package

ramonskie:
a new suggestion was made 
```
So a quick fix would be:
$ sudo touch /usr/share/vim/vim82/defaults.vim
```

PR #266, opened 2022-12-19
add vim-runtime to jammy

ramonskie:
this would solve #239


aramprice (commented):
This looks ok.

One nit I have though is that combining two unrelated fixes into a single PR is confusing, especially if the `vim-runtime` fix might be reverted in the future.

ramonskie:
i removed the commit related to the gitingore issue

rkoster (approved):

beyhan (approved):

rkoster:
Thanks! @ramonskie

PR #270, opened 2023-01-16
install ruby properly so we don't need sudo to install gems

ramonskie:


ci/docker/os-image-stemcell-builder-bionic/Dockerfile
```diff
@@ -24,25 +24,47 @@ RUN apt-get update \
...
-RUN apt-get install -y libxslt-dev libxml2-dev && apt-get clean
+# install ruby
+RUN \
+  wget -O ruby-install-0.8.3.tar.gz https://github.com/postmodern/ruby-install/archive/v0.8.3.tar.gz \
+  && tar -xzvf ruby-install-0.8.3.tar.gz \
+  && cd ruby-install-0.8.3/ \
+  && make install \
+  && cd .. \
+  && rm -r ruby-install* \
+  && ruby-install 3.0.2
```
beyhan: We install ruby now without providing a sha and the `--cleanup` flag as it is done in the install-ruby.sh [here](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/bb1b11507c495d453fe0fb4bf617b8b2aa31b3a6/ci/docker/os-image-stemcell-builder/install-ruby.sh#L34). Additionally, we don't install bundler any more like it was done [before](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/bb1b11507c495d453fe0fb4bf617b8b2aa31b3a6/ci/docker/os-image-stemcell-builder/install-ruby.sh#L40-L41). Is there a reason for these changes?

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
```diff
@@ -24,25 +24,47 @@ RUN apt-get update \
...
+  && cd .. \
+  && rm -r chruby-0.3.9* \
+  && echo 'source /usr/local/share/chruby/chruby.sh; chruby 3.0.2' > ~ubuntu/.bashrc
 
-ADD install-ruby.sh /tmp/install-ruby.sh
-RUN chmod a+x /tmp/install-ruby.sh
-RUN cd /tmp && ./install-ruby.sh && rm install-ruby.sh
+# ADD install-ruby.sh /tmp/install-ruby.sh
+# RUN chmod a+x /tmp/install-ruby.sh
+# RUN cd /tmp && ./install-ruby.sh && rm install-ruby.sh
```
beyhan: Should we remove the [install-ruby.sh](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/ci/docker/os-image-stemcell-builder/install-ruby.sh)?

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
```diff
@@ -24,25 +24,47 @@ RUN apt-get update \
...
 COPY --from=golang:1 /usr/local/go /usr/local/go
 ENV GOROOT=/usr/local/go PATH=/usr/local/go/bin:$PATH
 
 ADD scripts/update.sh /tmp/update.sh
 RUN /tmp/update.sh && rm /tmp/update.sh
 
-ENV OVF_TOOL_INSTALLER VMware-ovftool-4.1.0-2459827-lin.x86_64.bundle
-ENV OVF_TOOL_INSTALLER_SHA1 b907275c8d744bb54717d24bb8d414b19684fed4
+# ENV OVF_TOOL_INSTALLER VMware-ovftool-4.1.0-2459827-lin.x86_64.bundle
+# ENV OVF_TOOL_INSTALLER_SHA1 b907275c8d744bb54717d24bb8d414b19684fed4
```
beyhan: Why we need these changes in this commit?

beyhan (changes requested):

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
ramonskie: these can be cleanup yes

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
ramonskie: yes we should remove this

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
ramonskie: i took this way of installing ruby from the jammy stemcells.
- bundler is now installed by default with using these ruby installer

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
beyhan: Ok, let's remove it with this pr?

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
beyhan: Ok, let's clean this up with this pr

beyhan (dismissed):

ci/docker/os-image-stemcell-builder-bionic/Dockerfile
nouseforaname: Then I think we should add sha checking for both instead of dropping it in both places.

nouseforaname (approved):
lgtm. thanks

beyhan (approved):

PR #278, opened 2023-01-30
move to ruby 3.2

ramonskie:


klakin-pivotal:
Why are we doing this now? Given that Ruby 3.0 is still in active support, (see <https://www.ruby-lang.org/en/downloads/branches/>) I would expect Ruby 3.1 to be actively supported for at least the next several years. Why not wait for a while for other folks to find problems with Ruby 3.2 and get fixes in?

ramonskie:
i bumped it because bosh is now on ruby 3.2
and our intergration containers are also now running on ruby 3.2 https://github.com/cloudfoundry/bosh/blob/main/ci/dockerfiles/integration/Dockerfile

ci/docker/os-image-stemcell-builder-jammy/Dockerfile
```diff
@@ -57,14 +57,17 @@ RUN groupadd -o -g ${GROUP_ID} ubuntu                  \
   && echo 'ubuntu ALL=NOPASSWD:ALL' >> /etc/sudoers
 
 # we install a local ruby (3.1) because Jammy is lagging at 2.7
```
cunnie: This comment is stale; you can change to `we install a local Ruby (3.x) because Jammy is lagging at 2.7`.

Note: I probably wrote this comment originally, and I take full responsibility.

cunnie (changes requested):
Could you please update the git commit message to include what you replied to @klakin-pivotal ? e.g.

```
We update Ruby to 3.2:
- BOSH at Ruby 3.2
- containers running our integration tests are also at Ruby 3.2
```

cunnie (approved):
LGTM

ystros (approved):

PR #287, opened 2023-04-04
Add systemd config to prevent chronyd from locking bind mounts

jpalermo:
Issue: https://github.com/cloudfoundry/bosh/issues/2433

Chrony's systemd config uses ReadWritePaths to limit the files it has access to. When this is configured, systemd creates a file system namespace for the service. This namespace ties up the existing bind mounts, even if the service isn't given access to those particular paths. This can cause disk unmounting to fail if the file namspace was created when that mounting existed.

This was most commonly seen when chrony had been restarted AFTER the persistent disk had been mounted. Normally when doing a persistent disk migration.

This changes adds back InaccessiblePaths to prevent the file namspace from having access to /var/vcap/store.

This work was done previously in: #152 but was lost as part of a later cleanup.

### DIFF
--- a/bosh-stemcell/spec/support/os_image_systemd_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_systemd_shared_examples.rb
@@ -12,6 +12,11 @@ shared_examples_for 'a systemd-based OS image' do
       it { should be_enabled }
     end
 
+    describe file('/etc/systemd/system/chronyd.service.d/prevent_mount_locking.conf') do
+      it { should be_file }
+      its(:content) { should match /^InaccessiblePaths=-\/var\/vcap\/store/ }
+    end
+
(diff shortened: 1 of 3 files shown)


ramonskie:
do we also need to revert https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/65fdd0f8cd8468980664f3afea634810a0575959

jpalermo:
Reload seems better than restart. I saw some commits where we added the restart and the reason we did that was because reload was broken in Xenial. So the restart we were doing was a hack.

ramonskie (approved):
lgtm

cunnie (approved):
LGTM.

With the following caveats: I'm not a systemd wizard, and I assume that `InaccessiblePaths=-/var/vcap/store` does the right thing.

PR #288, opened 2023-04-17
Moving phc conf to chrony config

lnguyen:
Note that this PR is merely a file change—instead of being placed in the `sources.d/ ` subdirectory, the directive to use the reference clock is placed in the `conf.d/` subdirectory, for the directive to use the reference clock is **not** a source definition; rather, it's a configuration definition.

If placed in `source.d/`, the directive has no effect.
[#184816605] Fix chrony azure phc config and update Ops Manager documentation

### DIFF
--- a/stemcell_builder/stages/bosh_azure_chrony/apply.sh
+++ b/stemcell_builder/stages/bosh_azure_chrony/apply.sh
@@ -6,7 +6,7 @@ base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
 source $base_dir/lib/prelude_bosh.bash
 
-cat > $chroot/etc/chrony/sources.d/azure_ptp.sources <<EOF
+cat > $chroot/etc/chrony/conf.d/azure_ptp.conf <<EOF
 # created by $0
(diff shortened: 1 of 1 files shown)


cunnie (changes requested):
This is great! Thanks.

Could you please re-write the commit message to include something along the lines of the following:

> Note that this PR is merely a file change—instead of being placed in the `sources.d/` subdirectory, the directive to use the reference clock is placed in the `conf.d/` subdirectory. The directive doesn't work if it's placed in `sources.d` because it's not a source directive, it's a configuration directive.

lnguyen:
Updated @cunnie

jpalermo (approved):
👍

cunnie:
@lnguyen Oops, I meant could you please include that text in the _git_ commit message, not the _PR_ commit message. I always look at _git_, rarely at PRs. Thanks.

lnguyen:
ugh... i cant fix this commit message cause its on protected branch...

ramonskie:
ahh yes its protected due to `ubuntu-jammy\*` branch rule
so we either accept how it is for now.
or you need to rename the branch

lnguyen:
Moving PR to https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/289

PR #290, opened 2023-05-12
Fixing rsyslog for warden stemcells

lnguyen:
Warden stemcells uses runit and doesn't have systemd setup. Rsyslog doesn't work in warden stemcell with new jammy rsyslog conf, removing SystemLogSocketName fixes this issue.


ramonskie (approved):
lgtm

jpalermo (approved):

PR #291, opened 2023-05-24
Fix escaping for rsyslog command

ragaskar:


ragaskar:
woop wrong branch

ragaskar:
ok now it should be cool

PR #292, opened 2023-05-24
Fix chroot command

ragaskar:
- It'll definitely work this time.
- Necessary to escape the \ as well (which results in 3 \\\, the first two print a `\` and the latter one prints a `$`. This is because there are a few levels of bash interpolation occurring)


jpalermo (approved):
👍

jpalermo (approved):
👍

aramprice (approved):
lgtm

PR #297, opened 2023-08-17
Add new stage to disable RemoveIPC in systemd

cunnie:
When RemoveIPC is enabled, it cleans up shared files for the vcap user which can cause problems for long running processes such as Postgres that depend on these shared files

Drive-by: We cleaned-up the README: we assume we're building on x86_64, And also we use the pre-built Docker image instead of building our own.


aramprice (approved):
lgtm

jpalermo (approved):
👍

PR #302, opened 2023-11-13
remove go_agent in the stemcell name

ramonskie:
as the go agent is the only bosh agent installed. it makes no sense to add it in the file name.

with this change we will break some pipelines as they have filter that specifically states `go_agent`
example in the stemcell publisher pipeline.
https://github.com/cloudfoundry/bosh-stemcells-ci/blob/master/pipelines/publisher/pipeline.yml#L57

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,11 @@ The arguments to `stemcell:build_with_local_os_image` are:
 You can find the resulting stemcell in the `tmp/` directory of the host, or in
 the `/opt/bosh/tmp` directory in the Docker container. Using the above example,
 the stemcell would be at
-`tmp/bosh-stemcell-0.0.8-vsphere-esxi-ubuntu-jammy-go_agent.tgz`. You can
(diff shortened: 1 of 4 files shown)


rkoster (approved):

ramonskie:
bosh.io also seems to rely on the bosh_agent regex
see
- https://github.com/cloudfoundry/bosh-io-web/blob/main/stemcell/stemsrepo/s3_stemcell.go#L14
- https://github.com/cloudfoundry/bosh-io-web/blob/main/ui/stemcell/nav.go#L24

beyhan:
We need more investigations here e.g. in bosh.io to understand the real impact of this change.

selzoc:
Is this something we're still interested in doing?  It seems liable to break a lot of things with no clear benefit (other than brevity).

ramonskie:
can be closed as we only remove this for noble and future releases.

PR #303, opened 2023-11-20
Use fips preview packages

rkoster:


ramonskie:
at the moment we have a weird issue where usbmuxd user is added and a `usbmuxd,libusbmuxd6` packages are installed.
i have no idea where this is coming from as we only installed fips related packages. so why would a usbmux deamon be installed that is used for appl/ios devices
@toabctl @rthill91 any ideas?

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -1,29 +1,155 @@
...
+      its(:content) { should match %r{linux\t/boot/vmlinuz-\S+-fips root=UUID=\S* ro } }
+      its(:content) { should match %r{initrd\t/boot/initrd.img-\S+-fips} }
+    end
+  end
+
+  linux_version_regex = 's/linux-(.+)-([0-9]+).([0-9]+).([0-9]+)-([0-9]+)/linux-\1-\2.\3/'
+
+  describe 'installed packages' do
+    dpkg_list_packages = "dpkg --get-selections | cut -f1 | sed -E '#{linux_version_regex}'"
+    # TODO: maby we can use awk "dpkg --get-selections | awk '!/linux-(.+)-([0-9]+.+)/&&/linux/{print $1}'"
```
jpalermo, line 48: Is this something we want to do?

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -1,29 +1,155 @@
...
+    let(:dpkg_list_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy.txt')).map(&:chop) }
+    let(:dpkg_list_fips_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-fips.txt')).map(&:chop) }
+    let(:dpkg_list_google_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-google-additions.txt')).map(&:chop) }
+    let(:dpkg_list_vsphere_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-vsphere-additions.txt')).map(&:chop) }
+    let(:dpkg_list_azure_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-azure-additions.txt')).map(&:chop) }
+    let(:dpkg_list_cloudstack_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-cloudstack-additions.txt')).map(&:chop) }
+    let(:dpkg_list_softlayer_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-softlayer-additions.txt')).map(&:chop) }
+
+    describe command(dpkg_list_packages), {
+      exclude_on_fips: true,
```
jpalermo: Are these all supposed to be exclude_on_fips: true even though they’re in the fips spec?

bosh-stemcell/spec/stemcells/fips_spec.rb
rkoster: The problem being solved here is not failing our test on every kernel patch:
```
debug/9f7b9093-f357-4397-8a29-67d6d5e7bd56:~$ dpkg --get-selections | cut -f1 | grep linux
binutils-x86-64-linux-gnu
console-setup-linux
libselinux1:amd64
linux-base
linux-fips
linux-fips-headers-5.15.0-73
linux-firmware
linux-headers-5.15.0-73-fips
linux-headers-fips
linux-image-5.15.0-73-fips
linux-image-fips
linux-image-hmac-5.15.0-73-fips
linux-libc-dev:amd64
linux-modules-5.15.0-73-fips
linux-modules-extra-5.15.0-73-fips
```

Becomes:
```
dpkg --get-selections | cut -f1 | sed -E 's/linux-(.+)-([0-9]+).([0-9]+).([0-9]+)-([0-9]+)/linux-\1-\2.\3/' | grep linux
binutils-x86-64-linux-gnu
console-setup-linux
libselinux1:amd64
linux-base
linux-fips
linux-fips-headers-5.15
linux-firmware
linux-headers-5.15-fips
linux-headers-fips
linux-image-5.15-fips
linux-image-fips
linux-image-hmac-5.15-fips
linux-libc-dev:amd64
linux-modules-5.15-fips
linux-modules-extra-5.15-fips
util-linux
```

The proposed `awk` based solution (which can not be used as is because it filters the list down to only packages with linux in their name) proposes to drop versions from linux packages all together:
```
sudo dpkg --get-selections | awk '!/linux-(.+)-([0-9]+.+)/&&/linux/{print $1}'
binutils-x86-64-linux-gnu
console-setup-linux
libselinux1:amd64
linux-base
linux-fips
linux-firmware
linux-headers-fips
linux-image-fips
linux-libc-dev:amd64
util-linux
```

I'm not a fan of removing versions altogether and would like our tests to still fail on Minor kernel bumps. 
If someone feels like the current behavior could be implemented with `awk` in a better way, be my guest, but this should not block this PR.

bosh-stemcell/spec/stemcells/fips_spec.rb
rkoster: Lol, yeah that kinda would be the point of a fips spec 😄 Good catch!

bosh-stemcell/spec/stemcells/fips_spec.rb
ramonskie: yeah i wrote this `#todo` before we where solving it with the `update_kernel_static_libraries` function.
i think we can remove this todo comment it has been there for a few years now

bosh-stemcell/spec/stemcells/cis_spec.rb
```diff
@@ -49,9 +48,27 @@
...
+    exclude_on_google: true,
+    exclude_on_vcloud: true,
+    exclude_on_vsphere: true,
+    exclude_on_warden: true,
+    exclude_on_openstack: true,
+    exclude_on_cloudstack: true,
+    exclude_on_softlayer: true,
+  } do
+    it 'confirms that all CIS test cases ran' do
+      expect($cis_test_cases.to_a).to match_array(base_cis_test_cases + ['CIS-2.24', ])
```
beyhan: why we have comma after `'CIS-2.24'` but no value? Did we forget to add a CIS case here?

bosh-stemcell/spec/support/os_image_shared_examples.rb
```diff
@@ -760,7 +760,7 @@
 
   end
 
-  describe 'record use of privileged programs (CIS-8.1.12)', exclude_on_fips: false do # TODO: revert exlude to true when fips is enabled
+  describe 'record use of privileged programs (CIS-8.1.12)' do
```
beyhan, line 763: I think keeping the `cis_test_cases` full and use `exclude_on_fips: true` here to skip this is less error prone compared to take `CIS-8.1.12` out of that list.

cunnie (dismissed):
I agree with everything Beyhan says.

Looks good to me. I realize that the introduction of FIPS brings another dimension of complexity into an already-tangled codebase, but I don't see an easy solution.

rkoster:
All CIS-related exclusions were still from the bionic fips branch where we used an AWS-specific kernel. Now that we have switched to a generic fips kernel the CIS spec just works.

beyhan (approved):

cunnie (approved):
LGTM, with the usual caveats: FIPS adds another dimension of complexity to an already-complex build process, but I don't see an easy way of simplifying.

PR #305, opened 2023-12-08
use snapshot apt when building from tag

ramonskie:
when building from a tag the tag date wil be used for setting the date for the snapshot apt.

this will give us the ability to replicate stemcell builds
except for the rsyslog packages at the moment

### DIFF
--- a/bosh-stemcell/spec/os_image/ubuntu_jammy_spec.rb
+++ b/bosh-stemcell/spec/os_image/ubuntu_jammy_spec.rb
@@ -36,9 +36,9 @@ describe 'Ubuntu 22.04 OS image', os_image: true do
 
   describe 'base_apt' do
     describe file('/etc/apt/sources.list') do
-      its(:content) { should match 'deb http://archive.ubuntu.com/ubuntu jammy main universe multiverse' }
-      its(:content) { should match 'deb http://archive.ubuntu.com/ubuntu jammy-updates main universe multiverse' }
-      its(:content) { should match 'deb http://security.ubuntu.com/ubuntu jammy-security main universe multiverse' }
+      its(:content) { should match 'deb http:\/\/(archive|snapshot).ubuntu.com\/ubuntu(|\/\d*T\d*Z) jammy main universe multiverse' }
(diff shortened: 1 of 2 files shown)


lnguyen (approved):
LGTM

klakin-pivotal:
Do you have a link to documentation that describes Canoncial's commitments to keeping this service up and running, packages will be included in it, and who will get access to it?

All I can find is this <https://ubuntu.com/blog/ubuntu-snapshots-on-azure-ensuring-predictability-and-consistency-in-cloud-deployments>, which seems to suggest that it's closely entangled with Azure for some reason, and this old comment <https://askubuntu.com/questions/499871/is-there-ubuntus-analogue-of-snapshot-debian-org#comment1076354_499871>, which indicates that (historically) `snapshots.ubuntu.com` has been low-quality.

jpalermo:
Pondering an alternative approach that ensures the builds use the same timestamp for the snapshot repo both initially and when building at a later date.

- Concourse generates a new timestamp at the start of a stemcell build via a time resource `put` action. The value of that is then set as an environment variable in the format the snapshot repo expects
- The builder code checks for that environment variable, and if set, uses that for the snapshot repo.
- The builder code also checks for a `build_time` (or something) in the root of the repo, and if that exists it will fall back to it instead
- If neither exist, it uses the regular archive repos
- After a successful stemcell build, when we normally just tag the repo, we instead make a commit adding the `build_time` file to the root. This commit is tagged and pushed, but not to the branch. This leaves the tag with a `build_time` file, but keeps the `build_time` file from ever entering the main branches.

Considerations
- I believe we'll need to modify `bosh-stemcell/lib/bosh/stemcell/build_environment.rb` to ensure the environment variable with the timestamp is available to `base_apt/apply.sh`
- I've never pushed just a tag and not a branch, but the concourse resource does have a `only_tag` which sounds like it should work.

Thoughts?

jpalermo:
PRs for alternate approach:
https://github.com/cloudfoundry/bosh-stemcells-ci/pull/22
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/317

rkoster:
Closing because of PR's from @jpalermo

aramprice:
@ramonskie would you mind deleting this branch, thanks!

PR #306, opened 2023-12-11
Enable esm-apps for use with ESM on Xenial

ystros:
The ubuntu-advantage-tools CLI is also called `pro` now; while there's currently an alias for `ua`, that could change in the future, so we've pre-emptively started using the new name.

[#186639388] Update xenial stuff to use esm-apps in addition to esm-infra


aramprice (approved):
lgtm

jpalermo (approved):
👍

PR #307, opened 2023-12-19
Adding support for GPU variant

lnguyen:
Enhancing support for Nvidia GPU drivers installation on the stemcell is crucial to accommodate diverse GPU workloads across various Infrastructure-as-a-Service (IaaS) platforms. Recognizing the limitation of the original 5GB root disk, I've introduced configurability through environment variables, specifically DISK_SIZE, to address this constraint. Given the substantial size of Nvidia drivers, a larger root disk, currently set at 13GB, becomes essential for seamless installation.

Our exploration into different installation methods involved employing pre-start scripts that relocated multiple folders to /var/vcap/data before initiating the driver installation. While successful, this approach left us with a mere 200MB of available space, presenting suboptimal conditions. Operating with such minimal free space poses potential risks, including conflicts with antivirus software or other machine-specific runtime configurations.

It's important to note that this marks the initial steps toward optimizing the environment for GPU workloads. However, the long-term viability of this approach may be limited, considering the potential requirement for diverse drivers tailored to different GPU models.


jpalermo:
Could you expand on your description a bit @lnguyen 

Details about struggles with the root disk would be helpful too

bosh-stemcell/spec/stemcells/gpu_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+        end
+      end
+    end
+
+    context 'System command files must have mode 0755 or less permissive (stig: V-38469)' do
+        describe command('find -L /bin /usr/bin /usr/local/bin /sbin /usr/sbin /usr/local/sbin -perm /022 -type f') do
+          its (:stdout) { should eq('') }
+        end
+    end
+
```
klakin-pivotal, line 26: These tests appear to be word-for-word identical to tests in `stemcell_shared_examples.rb` and `os_image_shared_examples.rb`. Is there a reason why they've been duplicated here?

bosh-stemcell/spec/stemcells/gpu_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+_runit-log:!::
+admin:!::vcap
+vcap:!::syslog
+bosh_sshers:!::vcap
+bosh_sudoers:!::
+nvidia-persistenced:!::
+HERE
+        end
+    end
+end
```
klakin-pivotal, line 238: Because this stemcell variant is based on Jammy, is there any reasonable way to express the contents of ~these~ [the] user/group files [above] as additions to the file definitions in the Jammy test, rather than duplication of all of the content of those files?

bosh-stemcell/spec/stemcells/ubuntu_jammy_spec.rb
klakin-pivotal, line 1: What's up with all the `exclude_on_gpu` stuff in this file? Why are we not running these tests on the GPU stemcell variant?

bosh-stemcell/spec/stemcells/cis_spec.rb
```diff
@@ -49,7 +49,11 @@
   end
 
 
-  context "For all infrastructure except Azure and Cloudstack", {exclude_on_azure:true, exclude_on_cloudstack:true} do
+  context "For all infrastructure except Azure, Cloudstack and GPU", {
+    exclude_on_azure: true,
+    exclude_on_cloudstack: true,
+    exclude_on_gpu: true
```
klakin-pivotal, line 55: Why not run these tests on the GPU variant?

bosh-stemcell/spec/os_image/ubuntu_jammy_spec.rb
klakin-pivotal, line 1: I worry about the `exclude_on_gpu` annotations here... but see my comments on `gpu_spec.rb`.

jpalermo:
Ping @lnguyen for updates

lnguyen:
I've converted to a draft for now, I need to fix up tests

lnguyen:
delete for now

aramprice:
@lnguyen would you mind deleting this branch, thanks!

PR #308, opened 2023-12-22
Changes logrotate.timer trigger to OnCalendar

klakin-pivotal:
**NOTE**: This commit makes changes to the stemcell acceptance tests, but the commit author has not run those tests.

This commit changes the logrotate Timer Unit's scheduling mechanism from "OnUnitActiveSec" to "OnCalendar".

Why? A customer reported that on their Jammy-stemcell-based software, logrotate was never being run. Support reproduced the problem in their lab.

Investigation revealed that scheduling oneshot Timer Units with "OnUnitActiveSec" is something that has been long-known to be unsupported [1], and has had recent trouble reports from people running software on Jammy [2].

As further evidence of this change's likely correctness, Canonical also made a similar change for the same reason back in 2019. [3] (The programmer making the fix tries a few different things, but eventually discovers that "OnCalendar" works correctly. [4])

Unfortunately, this problem usually does not happen, and getting it TO happen can be very difficult, so proving the correctness of this fix is hard.

[1] <https://github.com/systemd/systemd/issues/6680#issuecomment-326228938>
[2] <https://github.com/systemd/systemd/issues/6680#issuecomment-1852339223>
[3] <https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1829968>
[4] <https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1829968/comments/5>

### DIFF
--- a/src/github.com/cloudfoundry/stemcell-acceptance-tests/ipv4director/smoke/smoke_test.go
+++ b/src/github.com/cloudfoundry/stemcell-acceptance-tests/ipv4director/smoke/smoke_test.go
@@ -29,7 +29,7 @@ func fillUpFileWithRandomCharacters(filepath string) {
 var _ = Describe("Stemcell", func() {
 	Context("when logrotate wtmp/btmp logs", func() {
 		It("should rotate the wtmp/btmp logs", func() {
-			stdOut, stdErr, exitStatus, err := bosh.Run("ssh", "default/0", `sudo sed -E -i "s/OnUnitActiveSec.*/OnUnitActiveSec=30s/" /etc/systemd/system/logrotate.timer && sudo sed -i "/RandomizedDelaySec/d" /etc/systemd/system/logrotate.timer && sudo systemctl daemon-reload`)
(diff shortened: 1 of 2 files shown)


jpalermo:
@ramonskie I added you as a reviewer since you were involved (although >2 years ago) with replacing cron with the systemd timer config.

Doubt you have any context on it still, but wanted to see if you remember anything.

jpalermo (dismissed):
👍

klakin-pivotal:
I have force-pushed changes (that I had forgotten) that update the comment in `logrotate.timer`. No actual changes to the already-reviewed functionality have been made.

lnguyen (approved):
LGTM

ramonskie:
i know why we did it. and it was to workaround a systemd bug.
if this change improves on that then im all for it.
as systemd is mostly a hellish experience

klakin-pivotal:
> as systemd is mostly a hellish experience

Yes. This is not the only example of the hell, but imagine thinking that it is okay to ship a `cron` replacement that can be configured to sometimes not run the task you scheduled. Madness.

> if this change improves on that then im all for it.

It probably does. The customer reports that doing this fixed this problem for them.
Would you be willing to Approve this PR so it can be merged and we can make a stemcell with it?

ramonskie (approved):

PR #309, opened 2024-01-05
enable the use of iaas-specific kernels

rthill91:
Use `UBUNTU_IAAS_KERNEL` environment variable to identify the use of, and label for, iaas specific kernels (e.g. `aws`)

This somewhat replicates how fips was handled in the bionic branch, but with an attempt to be more generic.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
@@ -63,7 +63,7 @@ module Bosh::Stemcell
         'UBUNTU_ISO' => environment['UBUNTU_ISO'],
         'UBUNTU_MIRROR' => environment['UBUNTU_MIRROR'],
         'UBUNTU_ADVANTAGE_TOKEN' => environment['UBUNTU_ADVANTAGE_TOKEN'],
-        'IAAS_KERNEL' => environment['IAAS_KERNEL'],
+        'UBUNTU_IAAS_KERNEL' => environment['UBUNTU_IAAS_KERNEL'],
       }
     end
 
--- a/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
@@ -63,6 +63,7 @@ module Bosh::Stemcell
         'UBUNTU_ISO' => environment['UBUNTU_ISO'],
         'UBUNTU_MIRROR' => environment['UBUNTU_MIRROR'],
         'UBUNTU_ADVANTAGE_TOKEN' => environment['UBUNTU_ADVANTAGE_TOKEN'],
+        'IAAS_KERNEL' => environment['IAAS_KERNEL'],
       }
     end
 
--- a/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
@@ -63,7 +63,7 @@ module Bosh::Stemcell
         'UBUNTU_ISO' => environment['UBUNTU_ISO'],
         'UBUNTU_MIRROR' => environment['UBUNTU_MIRROR'],
         'UBUNTU_ADVANTAGE_TOKEN' => environment['UBUNTU_ADVANTAGE_TOKEN'],
-        'UBUNTU_IAAS_KERNEL' => environment['UBUNTU_IAAS_KERNEL'],
+        'UBUNTU_FIPS_USE_IAAS_KERNEL' => environment['UBUNTU_FIPS_USE_IAAS_KERNEL'],
       }
     end
 
--- a/bosh-stemcell/spec/stemcells/fips_spec.rb
+++ b/bosh-stemcell/spec/stemcells/fips_spec.rb
@@ -1,9 +1,9 @@
 require 'spec_helper'
 
 describe 'FIPS Stemcell', os_image: true do
+  use_iaas_kernel = ENV.fetch('UBUNTU_FIPS_USE_IAAS_KERNEL', 'false') != 'true'
   context 'installed by system_kernel' do
     infrastructure = ENV['STEMCELL_INFRASTRUCTURE']
-    use_iaas_kernel = ENV['UBUNTU_FIPS_USE_IAAS_KERNEL']
     describe package(use_iaas_kernel ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
       it { should be_installed }
     end
@@ -39,7 +39,7 @@ describe 'FIPS Stemcell', os_image: true do
     describe file('/boot/grub/grub.cfg') do
       it { should be_file }
       its(:content) { should match %r{linux\t/boot/vmlinuz-\S+-fips root=UUID=\S* ro } }
-      if ENV.key?("UBUNTU_FIPS_USE_IAAS_KERNEL")
+      if use_iaas_kernel
         its(:content) { should match %r{initrd\t/boot/microcode.cpio /boot/initrd.img-\S+-fips} }
       else
         its(:content) { should match %r{initrd\t/boot/initrd.img-\S+-fips} }
@@ -61,6 +61,7 @@ describe 'FIPS Stemcell', os_image: true do
     let(:dpkg_list_azure_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-azure-additions.txt')).map(&:chop) }
     let(:dpkg_list_cloudstack_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-cloudstack-additions.txt')).map(&:chop) }
     let(:dpkg_list_softlayer_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-softlayer-additions.txt')).map(&:chop) }
+    let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
 
     describe command(dpkg_list_packages), {
       exclude_on_alicloud: true,
@@ -72,11 +73,8 @@ describe 'FIPS Stemcell', os_image: true do
       exclude_on_openstack: true,
       exclude_on_softlayer: true,
     } do
-      let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
-      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
-
       it 'contains only the base set of packages plus aws-specific kernel packages' do
-        skip "Test skipped due to non-aws kernel" if (use_iaas_kernel && infrastructure != 'aws')
+        skip "Test skipped due to generic kernel" unless use_iaas_kernel
         pkg_list = dpkg_list_ubuntu.concat(dpkg_list_aws_fips_ubuntu)
         pkg_list.delete('linux-firmware')
         pkg_list.delete('wireless-regdb')
@@ -92,8 +90,6 @@ describe 'FIPS Stemcell', os_image: true do
       exclude_on_azure: true,
       exclude_on_softlayer: true,
     } do
-      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
-
       it 'contains only the base set of packages for alicloud, aws, openstack, warden' do
         skip "Test skipped due to IAAS-specific kernel" if use_iaas_kernel
         expect(subject.stdout.split("\n")).to match_array(dpkg_list_ubuntu.concat(dpkg_list_fips_ubuntu))
--- a/bosh-stemcell/spec/stemcells/fips_spec.rb
+++ b/bosh-stemcell/spec/stemcells/fips_spec.rb
@@ -2,11 +2,9 @@ require 'spec_helper'
 
 describe 'FIPS Stemcell', os_image: true do
   context 'installed by system_kernel' do
-    iaas = ENV['UBUNTU_IAAS_KERNEL'] || ''
-    unless iaas.empty?
-      iaas = "#{iaas}-"
-    end
-    describe package("linux-image-#{iaas}fips") do
+    infrastructure = ENV['STEMCELL_INFRASTRUCTURE']
+    use_iaas_kernel = ENV['UBUNTU_FIPS_USE_IAAS_KERNEL']
+    describe package(use_iaas_kernel ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
       it { should be_installed }
     end
     describe package('linux-generic') do
@@ -41,7 +39,7 @@ describe 'FIPS Stemcell', os_image: true do
     describe file('/boot/grub/grub.cfg') do
       it { should be_file }
       its(:content) { should match %r{linux\t/boot/vmlinuz-\S+-fips root=UUID=\S* ro } }
-      if ENV.key?("UBUNTU_IAAS_KERNEL")
+      if ENV.key?("UBUNTU_FIPS_USE_IAAS_KERNEL")
         its(:content) { should match %r{initrd\t/boot/microcode.cpio /boot/initrd.img-\S+-fips} }
       else
         its(:content) { should match %r{initrd\t/boot/initrd.img-\S+-fips} }
@@ -74,8 +72,11 @@ describe 'FIPS Stemcell', os_image: true do
       exclude_on_openstack: true,
       exclude_on_softlayer: true,
     } do
+      let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
+      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
+
       it 'contains only the base set of packages plus aws-specific kernel packages' do
-        skip "Test skipped due to non-aws kernel" if (ENV['UBUNTU_IAAS_KERNEL'] != 'aws')
+        skip "Test skipped due to non-aws kernel" if (use_iaas_kernel && infrastructure != 'aws')
         pkg_list = dpkg_list_ubuntu.concat(dpkg_list_aws_fips_ubuntu)
         pkg_list.delete('linux-firmware')
         pkg_list.delete('wireless-regdb')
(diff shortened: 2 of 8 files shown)


bosh-stemcell/lib/bosh/stemcell/builder_options.rb
```diff
@@ -63,6 +63,7 @@ def environment_variables
         'UBUNTU_ISO' => environment['UBUNTU_ISO'],
         'UBUNTU_MIRROR' => environment['UBUNTU_MIRROR'],
         'UBUNTU_ADVANTAGE_TOKEN' => environment['UBUNTU_ADVANTAGE_TOKEN'],
+        'IAAS_KERNEL' => environment['IAAS_KERNEL'],
```
toabctl: nit: should this be `UBUNTU_IAAS_KERNEL` given that this is Ubuntu specific?

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -2,7 +2,11 @@
 
 describe 'FIPS Stemcell', os_image: true do
   context 'installed by system_kernel' do
-    describe package('linux-image-fips') do
+    iaas = ENV['IAAS_KERNEL'] || ''
+    unless iaas.empty?
```
toabctl: is this check for empty required? isn't a check for `ENV['IAAS_KERNEL']` enough? at least that's what's done in other places in this PR.

bosh-stemcell/lib/bosh/stemcell/builder_options.rb
toabctl: would be nice to check that this var can only be set to `aws` currently.

bosh-stemcell/spec/stemcells/fips_spec.rb
rthill91: I need to append that hyphen if `IAAS_KERNEL` isn't empty otherwise I want an empty string.

I could drop this in favor of the following oneliner
```rb
iaas = ("#{ENV['IAAS_KERNEL']}-" if ENV['IAAS_KERNEL']) || ''
```

ruby is also not my forte, so maybe I'm missing a more obvious way to do this.

bosh-stemcell/lib/bosh/stemcell/builder_options.rb
rthill91: renamed in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/309/commits/fed91af0a965655cfb6efa15085874f34de66863 and added a check for the value in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/309/commits/21485472676b68adf8bb711443abff0c92824d1c

rkoster:
@toabctl Is this PR ready for review? asking because there currently still is active discussion.

bosh-stemcell/spec/stemcells/fips_spec.rb
toabctl: thx (looks like I can't "resolve" this thread). Can somebody mark it as resolved please?

toabctl (approved):

stemcell_builder/stages/system_fips_kernel/apply.sh
```diff
@@ -14,6 +14,10 @@ add_on_exit "umount $chroot/sys"
 mock_grub_probe
 ua_attach
 ua_enable_fips
-pkg_mgr install fips-initramfs linux-fips
+kernel=linux-fips
+if [ ! -z "$UBUNTU_IAAS_KERNEL" ]; then
+    kernel=linux-$UBUNTU_IAAS_KERNEL-fips
+fi
+pkg_mgr install fips-initramfs $kernel
```
rkoster, line 21: Do we still need to specify the `cloud_type` if we install iaas specific kernel here anyway? I believe that code was only needed in the past when we relied on ua attach to install the kernel. But now that ua attach [runs without apt](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/309/files#diff-4741572db64c272a4d3237f9fbaeda98a4788cee27e1e1a1024c31bab6752b25R53-R62), this should no longer be needed.

bosh-stemcell/lib/bosh/stemcell/builder_options.rb
```diff
@@ -63,6 +63,7 @@ def environment_variables
         'UBUNTU_ISO' => environment['UBUNTU_ISO'],
         'UBUNTU_MIRROR' => environment['UBUNTU_MIRROR'],
         'UBUNTU_ADVANTAGE_TOKEN' => environment['UBUNTU_ADVANTAGE_TOKEN'],
+        'UBUNTU_IAAS_KERNEL' => environment['UBUNTU_IAAS_KERNEL'],
```
rkoster: To make this future proof I would like to propose change the variable name to: `UBUNTU_USE_IAAS_KERNEL` when set would install the iaas specific kernel for the given `$stemcell_infrastructure` (this variable is already available and gets used for example [here](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/stages/image_install_grub/apply.sh#L78C9-L85)).

Even more correct would be to use `UBUNTU_FIPS_USE_IAAS_KERNEL` because that's the current scope of the variable.

rkoster: To be able to use the `infrastructure.name` in the specs you would have to extend the rspec invocation [here](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/bosh-stemcell/lib/bosh/stemcell/build_environment.rb#L52-L56) to something like:
```ruby
      cmd = [
        "cd #{STEMCELL_SPECS_DIR};",
        "STEMCELL_IMAGE=#{image_file_path}",
        "STEMCELL_WORKDIR=#{work_path}",
        "STEMCELL_INFRASTRUCTURE=#{infrastructure.name}",  # ← added
        "OS_NAME=#{operating_system.name}",
        "OS_VERSION=#{operating_system.version}",
        "CANDIDATE_BUILD_NUMBER=#{@version}",
        "bundle exec rspec -fd#{exclude_exclusions}",
        "spec/os_image/#{operating_system_spec_name}_spec.rb",
        "spec/stemcells/#{operating_system_spec_name}_spec.rb",
        'spec/stemcells/go_agent_spec.rb',
        "spec/stemcells/#{infrastructure.name}_spec.rb",
        'spec/stemcells/stig_spec.rb',
        'spec/stemcells/cis_spec.rb'
      ]
```

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -2,7 +2,11 @@
...
 describe 'FIPS Stemcell', os_image: true do
   context 'installed by system_kernel' do
-    describe package('linux-image-fips') do
+    iaas = ENV['UBUNTU_IAAS_KERNEL'] || ''
+    unless iaas.empty?
+      iaas = "#{iaas}-"
+    end
+    describe package("linux-image-#{iaas}fips") do
       it { should be_installed }
     end
```
rkoster, line 9: I would use something like:
```ruby
let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
describe package(use_iaas_kernel ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
  it { should be_installed }
end
```

It could be a one-liner but given the environment variable name lengths, something like the above would be more readable.

rkoster (changes requested):
Would like this PR to refactored a bit so that the target infrastructure does not have to be specified twice on rake invocation, by making the introduced variable only toggle and not use it for passing down the target infrastructure into the tests.

stemcell_builder/stages/system_fips_kernel/apply.sh
rthill91: I think you're correct. We should be able to drop the client config overwrite altogether.

bosh-stemcell/spec/stemcells/fips_spec.rb
rthill91: I'm getting errors trying to use those variables inside describe. Specifically
> `use_iaas_kernel` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

Is there a scope concern keeping these as instance variables defined within this test? e.g.

```ruby
context 'installed by system_kernel' do
  infrastructure = ENV['STEMCELL_INFRASTRUCTURE']
  use_iaas_kernel = ENV['UBUNTU_FIPS_USE_IAAS_KERNEL']
  
  describe package(use_iaas_kernel ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
    it { should be_installed }
  end
  ...
end
```

rkoster: Yeah, using normal variables seems fine. I did not test my example, was just meant as an illustration.

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -49,12 +55,35 @@
 
     let(:dpkg_list_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy.txt')).map(&:chop) }
     let(:dpkg_list_fips_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-fips.txt')).map(&:chop) }
+    let(:dpkg_list_aws_fips_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-aws-fips.txt')).map(&:chop) }
     let(:dpkg_list_google_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-google-additions.txt')).map(&:chop) }
     let(:dpkg_list_vsphere_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-vsphere-additions.txt')).map(&:chop) }
     let(:dpkg_list_azure_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-azure-additions.txt')).map(&:chop) }
     let(:dpkg_list_cloudstack_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-cloudstack-additions.txt')).map(&:chop) }
     let(:dpkg_list_softlayer_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-softlayer-additions.txt')).map(&:chop) }
```
rkoster, line 65: I would move infrastructure and use_iaas_kernel here, so you do not have to duplicate them in nested describes.
```
      let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
```

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -49,12 +55,35 @@
...
+      exclude_on_warden: true,
+      exclude_on_azure: true,
+      exclude_on_openstack: true,
+      exclude_on_softlayer: true,
+    } do
+      let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
+      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
+
+      it 'contains only the base set of packages plus aws-specific kernel packages' do
+        skip "Test skipped due to non-aws kernel" if (use_iaas_kernel && infrastructure != 'aws')
```
rkoster: Could this line be changed to:
```
skip "Test skipped due to generic kernel" unless use_iaas_kernel
```
The infrastructure check should not be needed because the excludes.

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -37,7 +39,11 @@
     describe file('/boot/grub/grub.cfg') do
       it { should be_file }
       its(:content) { should match %r{linux\t/boot/vmlinuz-\S+-fips root=UUID=\S* ro } }
-      its(:content) { should match %r{initrd\t/boot/initrd.img-\S+-fips} }
+      if ENV.key?("UBUNTU_FIPS_USE_IAAS_KERNEL")
```
rkoster: Have you checked if the else can be reached? It could be that this environment variable always will be set, just to an empty string. Because of: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/04e622ab133a1d88a4a8b4b4685498925372c33e/bosh-stemcell/lib/bosh/stemcell/builder_options.rb#L66

rthill91: sorry I had to put this down for a couple weeks I've added a check for empty string as well as I think you're right about all the envvars being 'set' regardless

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -63,7 +91,10 @@
       exclude_on_azure: true,
       exclude_on_softlayer: true,
     } do
+      let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] }
```
rkoster: This line should not be needed because `use_iaas_kernel` is already defined here: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/309/files#diff-f18a6346d0ba4575fc99197220088b78b3e21a9696f0c5949166c0bb52463744R65

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -49,11 +55,33 @@
...
     let(:dpkg_list_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy.txt')).map(&:chop) }
     let(:dpkg_list_fips_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-fips.txt')).map(&:chop) }
+    let(:dpkg_list_aws_fips_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-aws-fips.txt')).map(&:chop) }
     let(:dpkg_list_google_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-google-additions.txt')).map(&:chop) }
     let(:dpkg_list_vsphere_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-vsphere-additions.txt')).map(&:chop) }
     let(:dpkg_list_azure_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-azure-additions.txt')).map(&:chop) }
     let(:dpkg_list_cloudstack_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-cloudstack-additions.txt')).map(&:chop) }
     let(:dpkg_list_softlayer_ubuntu) { File.readlines(spec_asset('dpkg-list-ubuntu-jammy-softlayer-additions.txt')).map(&:chop) }
+    let(:infrastructure) { ENV['STEMCELL_INFRASTRUCTURE'] }
+    let(:use_iaas_kernel) { ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] and ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] != "" }
```
rkoster: This could be written as:
```ruby
let(:use_iaas_kernel) { ENV.fetch('UBUNTU_FIPS_USE_IAAS_KERNEL', 'false') == 'true' }
```

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -2,7 +2,9 @@
 
 describe 'FIPS Stemcell', os_image: true do
   context 'installed by system_kernel' do
-    describe package('linux-image-fips') do
+    infrastructure = ENV['STEMCELL_INFRASTRUCTURE']
+    use_iaas_kernel = ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] and ENV['UBUNTU_FIPS_USE_IAAS_KERNEL'] != ""
```
rkoster:
```
use_iaas_kernel = ENV.fetch('UBUNTU_FIPS_USE_IAAS_KERNEL', 'false') == 'true'
```

rkoster (changes requested):
Almost there, just some small nitpicks.

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -3,8 +3,8 @@
 describe 'FIPS Stemcell', os_image: true do
   context 'installed by system_kernel' do
     infrastructure = ENV['STEMCELL_INFRASTRUCTURE']
-    use_iaas_kernel = ENV['UBUNTU_FIPS_USE_IAAS_KERNEL']
-    describe package(use_iaas_kernel ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
+    use_iaas_kernel = ENV.fetch('UBUNTU_FIPS_USE_IAAS_KERNEL', 'false') != 'true'
+    describe package(use_iaas_kernel != "" ? "linux-image-#{infrastructure}-fips" : "linux-image-fips") do
```
rkoster: `use_iaas_kernel` already is a boolean, no need to check against empty string

bosh-stemcell/spec/stemcells/fips_spec.rb
```diff
@@ -39,7 +39,7 @@
     describe file('/boot/grub/grub.cfg') do
       it { should be_file }
       its(:content) { should match %r{linux\t/boot/vmlinuz-\S+-fips root=UUID=\S* ro } }
-      if ENV.key?("UBUNTU_FIPS_USE_IAAS_KERNEL")
+      if ENV.key?("UBUNTU_FIPS_USE_IAAS_KERNEL") and ENV["UBUNTU_FIPS_USE_IAAS_KERNEL"] != ""
```
rkoster: The `use_iaas_kernel` let from further down [below](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/309/commits/6582e8d911973925a0fd354f2fa5f44adc4cfe6a#diff-f18a6346d0ba4575fc99197220088b78b3e21a9696f0c5949166c0bb52463744R65) should be moved up so that it can be reused here.

rkoster (approved):
LGTM :shipit:

ramonskie (approved):
looks good

rkoster:
Thanks! @rthill91

PR #311, opened 2024-01-10
fix: don't change the systemd default from shared

### LINKED ISSUE #310: Resize of persistent disks fails on recent Jammy stemcells - systemd mount namespacing issue?
We are experiencing the same symptoms as https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/195

Namely, resize of a persistent disk is failing with message:

```
Remounting persistent disk as readonly: Shelling out to mount: Running command: 'mount /dev/sdc1 /var/vcap/store -o ro', stdout: '', stderr: 'mount: /var/vcap/store: /dev/sdc1 already mounted or mount point busy
```

This is using the following vSphere stemcell: `bosh-vsphere-esxi-ubuntu-jammy-go_agent/1.329`

Outside of a resize operation, we can reproduce this on a BOSH VM by SSHing to it, and, in a root shell, running the commands:

```bash
# restart either auditd or chronyd
service auditd restart

# stop all bosh services
monit stop all
sleep 10 # or wait for all services to stop

# unmount persistent disk
umount /var/vcap/store

# try to remount it
mount /dev/sdc1 /var/vcap/store -o ro
# Output (ERROR):
# mount: /var/vcap/store: /dev/sdc1 already mounted or mount point busy.

# CLEANUP:
# mount as read-write - this will succeed
mount /dev/sdc1 /var/vcap/store
monit start all
```

(credit to @bg-govau for finding the above steps)

A more generic way to reproduce this, that does not require stopping BOSH services or messing with the real persistent disk, is as follows:

```bash
# create a temp file for our block device
tmpfile=$(mktemp)

# needs to be a decent size to format as XFS
truncate -s 50M "${tmpfile}"

# make a loop devices that points to this file
loopdev="$(losetup --find --show "${tmpfile}")"

# format it as XFS
mkfs -t xfs "${loopdev}"

# make somewhere to mount it
mntdir="$(mktemp -d)"

# mount it as read-write
mount "${loopdev}" "${mntdir}"

# restart auditd
service auditd restart

# now unmount it
umount "${mntdir}"

# now try to re-mount it, but as read-only (rw will work)
mount "${loopdev}" "${mntdir}" -o ro
# Output (ERROR):
# mount: /tmp/tmp.PMLRvWX6wP: /dev/loop0 already mounted or mount point busy.

# CLEANUP:
service auditd restart
losetup --detach "${loopdev}"
rmdir "${mntdir}"
rm "${tmpfile}"
```

However try as I may, I am *unable* to reproduce that 2nd method on an AWS standard Ubuntu Jammy VM - and I can't tell why! But it does happen every time on our stemcells, hence raising the issue here (rather than with `systemd`).

Here's what we've noted so far:

# `auditd` and `chronyd` have their own `mount_namespace`

If we run `pstree -S` we see that some services started by `systemd` have their own `mnt` (mount) namespace:

```
systemd─┬─VGAuthService
        ├─agetty
        ├─auditd(mnt)─┬─audisp-syslog
        │             └─2*[{auditd}]
        ├─chronyd(mnt,uts)───chronyd
        ├─cron
        ├─dbus-daemon
        ├─networkd-dispat
        ├─rsyslogd───10*[{rsyslogd}]
        ├─runsvdir─┬─runsv─┬─bosh-agent───11*[{bosh-agent}]
        │          │       └─svlogd
        │          └─runsv─┬─monit-actual───{monit-actual}
        │                  └─svlogd
        ├─sshd───sshd───sshd───bash───sudo───bash───pstree
        ├─systemd───(sd-pam)
        ├─systemd-journal
        ├─systemd-logind(mnt,uts)
        ├─systemd-network(mnt)
        ├─systemd-udevd(mnt,uts)
        └─vmtoolsd───3*[{vmtoolsd}]
```

# Mount propagation does not appear to be working properly

If these services are started *before* the persistent disk is mounted, then there are no issues. However if they are restarted *after* `/var/vcap/store` is mounted, then they can see those, and can continue to see those after the device is unmounted.

For example:

```
monit stop all
# wait for all to stop

# restart auditd to demonstrate the issue
service auditd restart

# verify it is currently mounted:
ls -l /var/vcap/store
# Output:
# total 0
# drwxrwxr-x 3 vcap vcap 18 Sep 28 01:23 mongod
# drwxr-xr-x 2 root root  6 Sep 28 01:23 mongodb_exporter
# drwxr-xr-x 2 root root  6 Sep 28 01:23 node_exporter

# unmount it
umount /var/vcap/store

# try to mount as read-only
mount /dev/sdc1 /var/vcap/store -o ro
# Output: 
# mount: /var/vcap/store: /dev/sdc1 already mounted or mount point busy.

# verify that it isn't mounted:
ls -l /var/vcap/store
# Output:
# total 0

# use nsenter to demonstrate that `auditd` still sees this as mounted in its own namespace:
nsenter --target $(cat /run/auditd.pid) --all ls -l /var/vcap/store
# Output:
# total 0
# drwxrwxr-x 3 vcap vcap 18 Sep 28 01:23 mongod
# drwxr-xr-x 2 root root  6 Sep 28 01:23 mongodb_exporter
# drwxr-xr-x 2 root root  6 Sep 28 01:23 node_exporter

# restart auditd again, it now forgets about the old mount
service auditd restart

# and we can now successfully mount it as read-only
mount /dev/sdc1 /var/vcap/store -o ro
```

And this explain why we are affected. One of our BOSH jobs has a pre-start script which modifies the `auditd` config, and restarts it.

# The `systemd` documentation is really confusing

The configuration for `auditd` (which is standard) includes:

```
### Security Settings ###
MemoryDenyWriteExecute=true
LockPersonality=true
ProtectControlGroups=true
ProtectKernelModules=true
ProtectHome=true
RestrictRealtime=true
```

See <https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html> for docs on these, some of which imply that those options will disable mount propagation.

If we comment them all out, then the issue clears up. And interestingly upstream `auditd` has removed some of these (2 years ago!) but those are not backported into Jammy.

However this isn't the full story. If we create a VM in AWS, using their standard Jammy image (`ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20231207`), and run those same reproduction steps above (the generic version), the mount propagation *does* work as expected (`umount` in host propagates into slave `auditd` process).

Therefore I suspect that this should work with the standard security options, however perhaps something else in the stemcell is undoing this?

# What is different between the vSphere stemcell and the AWS Jammy image?

Not much as far as I can tell. The versions of `auditd` and `systemd` are the same.

When I print out all config options `systemctl show --all` the only difference I noted that looked significant is that the BOSH one shows:

```
Tainted=cgroupsv1
```

And AWS did not. This felt like a solid lead, and I found:

https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/c3e6a15dc4e0a9d2ae65b871947a53bf1801b0ac/stemcell_builder/stages/image_install_grub/apply.sh#L87-L89

I tried setting the same in my AWS VM and restarted. After doing so it also showed `Tainted=cgroupsv1` however the issue still wouldn't reproduce.

I can't easily rebuild the stemcell (without this setting) to test from the other direction.

What I have noted is that the output of `mountinfo` differs on each system. On the vSphere BOSH system we see:

```bash
cat /proc/$(cat /run/auditd.pid)/mountinfo | grep "${loopdev}"

# Output:
# 475 474 7:0 / /tmp/tmp.3V7N9qPucV rw,relatime shared:211 - xfs /dev/loop0 rw,attr2,inode64,logbufs=8,logbsize=32k,noquota
```

And on AWS Jammy:

```bash
cat /proc/$(cat /run/auditd.pid)/mountinfo | grep "${loopdev}"

# Output:
# 524 194 7:8 / /tmp/tmp.jAtl1twL1v rw,relatime shared:337 master:346 - xfs /dev/loop8 rw,attr2,inode64,logbufs=8,logbsize=32k,noquota
```

Note that vSphere shows: `shared:211` and AWS shows `shared:337 master:346`.

I think that means the AWS one will receive updates from `master:346` whereas the vSphere one will not receive updates from anything - and I think that matches what we are seeing. However I can't pick where the `systemd` config differs.

And that's as far as we've got. Does anyone have an idea what is going on to make our systemd behave differently to the AWS Jammy one?

ghost:
If the default is private, then if namespaced services (like auditd/chronyd) are restarted after the persistent disk is mounted, then it is not possible to fully unmount that disk until those services are stopped. The system will appear to let you do so, however the device remains busy and subsequent attempts to remount (as read-only) will fail as the device is busy.

Fixes #310

There's probably a test that could be written to confirm it this is not added, however I don't know the DSL that Ruby is using for that and probably won't pick up for this one change.


ghost:
> CLA

This is a trivial change, and getting an authorized person to sign a CLA on my side will likely be onerous. Can this please be accepted as trivial and not requiring a CLA, or can a maintainer kindly reimplement the relevant deletion?

jpalermo:
CLA is unfortunately required and I don't think there is much flexibility on that. Could you do this in your "free time" and sign the CLA just as yourself?

Bionic is also not built from the jammy branch, so the bionic guard shouldn't be needed.

ghost:
Have updated the PR to drop the bionic guard, and it's now a delete only PR.

ghost:
CLA is now signed.

jpalermo (approved):
I was able to reproduce the issue on vsphere and gcp stemcells by deploying an empty VM with a persistent disk, ssh in and run `sudo service auditd restart`, then resize the persistent disk.

I was able to build a test stemcell with the make_root_dir_rprivate stage removed and then everything worked fine.

The removal is a bit concerning since the [commit that added it](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/914b37862104cad6b49b3630be4570c90ef6be20) didn't give a lot of details around the problem it was solving.

I was able to deploy Cloud Foundry with the test stemcell and pushed an app which all seemed to work fine.

ramonskie:
@jpalermo would it be a good idea to let the garden team look at it as well. as they probably have more test scenarios

rkoster (approved):

beyhan:
@ae-govau Thanks for the contribution!

bgandon:
Just a quick testimony: with the Stemcell v1.351 that ships this PR, the [blocking issues I had with Concourse on Bosh-Lite](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/310#issuecomment-1898893917) seems to be fixed.

Thank you all 👏

PR #314, opened 2024-01-17
change path lisubcmd.a in older fips kernel

ramonskie:
since kernel 6.5 the objtool/libsubcmd.a has been moved to objtool/libsubcmd/libsubcmd.a
fips still uses the 5.9 kernel, so we need to rewrite the static_libraries_list file to the old path

### DIFF
--- a/stemcell_builder/lib/prelude_apply.bash
+++ b/stemcell_builder/lib/prelude_apply.bash
@@ -41,9 +41,16 @@ function pkg_exists {
 }
 
 function update_kernel_static_libraries {
-    kernel_suffix=${1}
-    major_kernel_version=${2}
+  kernel_suffix=${1}
+  major_kernel_version=${2}
 
-    kernel_version=$(find $chroot/usr/src/ -name "linux-headers-$major_kernel_version.*$kernel_suffix" | grep -o "[0-9].*-[0-9]*$kernel_suffix")
-sed -i "s/__KERNEL_VERSION__/$kernel_version/g" $chroot/var/vcap/bosh/etc/static_libraries_list
+  kernel_version=$(find $chroot/usr/src/ -name "linux-headers-$major_kernel_version.*$kernel_suffix" | grep -o "[0-9].*-[0-9]*$kernel_suffix")
+  sed -i "s/__KERNEL_VERSION__/$kernel_version/g" $chroot/var/vcap/bosh/etc/static_libraries_list
+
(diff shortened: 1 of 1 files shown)


stemcell_builder/lib/prelude_apply.bash
```diff
@@ -41,9 +41,16 @@ function pkg_exists {
...
-    kernel_version=$(find $chroot/usr/src/ -name "linux-headers-$major_kernel_version.*$kernel_suffix" | grep -o "[0-9].*-[0-9]*$kernel_suffix")
-sed -i "s/__KERNEL_VERSION__/$kernel_version/g" $chroot/var/vcap/bosh/etc/static_libraries_list
+  kernel_version=$(find $chroot/usr/src/ -name "linux-headers-$major_kernel_version.*$kernel_suffix" | grep -o "[0-9].*-[0-9]*$kernel_suffix")
+  sed -i "s/__KERNEL_VERSION__/$kernel_version/g" $chroot/var/vcap/bosh/etc/static_libraries_list
+
+  # since kernel 6.5 the objtool/libsubcmd.a has been moved to objtool/libsubcmd/libsubcmd.a
+  # fips still uses the 5.9 kernel, so we need to rewrite the static_libraries_list file to the old path
+  # TODO: puting a todo tag here as this could probably be removed in the future if the kernel version gets bumped
+  if [[ "${stemcell_operating_system_variant}" == 'fips' ]]; then
+    sed -i "s/objtool\/libsubcmd\/libsubcmd.a/objtool\/libsubcmd.a/g" $chroot/var/vcap/bosh/etc/static_libraries_list
```
bgandon, line 54: Could we use the `sed -i "s|search|replacement|g" ...` form, so that forward slashes `/` don’t need to be escaped here?

bgandon (changes requested):

stemcell_builder/lib/prelude_apply.bash
ramonskie: i will fix this. i need to transfer this pr to another pr
as the branch structure i used falss under the branch protection rules :(

ramonskie:
fixed in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/315

aramprice:
@ramonskie would you mind deleting this branch, thanks!

ramonskie:
Its protected.. @rkoster could you remove this with active god mode

beyhan:
Even god mode can't delete the branch :-). Current branch protection rules are `ubuntu-*/*` and this doesn't allow deletion anyone. I renamed the branch and then deleted it.

PR #315, opened 2024-01-18
change path lisubcmd.a in older fips kernel

ramonskie:
since kernel 6.5 the objtool/libsubcmd.a has been moved to objtool/libsubcmd/libsubcmd.a
fips still uses the 5.9 kernel, so we need to rewrite the static_libraries_list file to the old path


beyhan:
Old pr: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/314

rkoster (approved):

bgandon (approved):

PR #317, opened 2024-01-29
Build Time can be passed via ENV variable or stored in root to limit snapshot apt repo packages

jpalermo:
See https://github.com/cloudfoundry/bosh-stemcells-ci/pull/22

Add build_time which can be passed in via env variable, or build_time.txt in the root of the repo.

Use that timestamp for the snapshot apt repo configuration.


ramonskie (approved):
it works perfectly :)

rkoster (approved):

PR #319, opened 2024-02-16
Add usg to jammy-fips builds

rthill91:
while not strictly required for fips, there is high overlap between users of fips images and various hardening profiles provided by usg

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy-fips.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy-fips.txt
@@ -1,7 +1,9 @@
+bsdmainutils
 fips-initramfs
 kcapi-tools
 libgcrypt20-hmac:amd64
 libkcapi1:amd64
+libopenscap8
 linux-fips
 linux-fips-headers-5.15
 linux-headers-5.15-fips
@@ -11,4 +13,7 @@ linux-image-fips
 linux-image-hmac-5.15-fips
 linux-modules-5.15-fips
 linux-modules-extra-5.15-fips
+ncal
 openssl-fips-module-3:amd64
+usg
+usg-benchmarks-1
--- a/stemcell_builder/lib/prelude_fips.bash
+++ b/stemcell_builder/lib/prelude_fips.bash
@@ -40,6 +40,10 @@ function ua_enable_fips() {
     run_in_chroot_without_apt ${chroot} "ua enable --assume-yes fips-preview"
 }
 
+function ua_enable_usg() {
+    run_in_chroot_without_apt ${chroot} "ua enable --assume-yes usg"
+}
+
 function run_in_chroot_without_apt() {
     local chroot=${1}
     local script=${2}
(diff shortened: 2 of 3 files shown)


rkoster:
@rthill91 could you provide a bit more context on the `usg bsdmainutils libopenscap8 ncal usg usg-benchmarks-1` packages? Will these packages cause breaking changes? I'm asking because cf-deployment validation against the fips-jammy stemcell has passed all, smoke and acceptance tests (as shared in yesterday's TOC meeting). If introducing these packages could make those tests fail again, we would at least need to discuss with the deployments working group how they would like to handle this.

rthill91:
> @rthill91 could you provide a bit more context on the `usg bsdmainutils libopenscap8 ncal usg usg-benchmarks-1` packages? Will these packages cause breaking changes? I'm asking because cf-deployment validation against the fips-jammy stemcell has passed all, smoke and acceptance tests (as shared in yesterday's TOC meeting). If introducing these packages could make those tests fail again, we would at least need to discuss with the deployments working group how they would like to handle this.

I would not expect any breaking changes from these dependencies.

- `USG` (The Ubuntu Security Guide) is the only new top-level dependency added here and serves as a way to check, apply, and validate various hardening profiles.
- `usg-benchmarks-1` contains the supported SCAP standards that usg can validate against and attempt to apply out of the box (user action required, nothing automated upon installation)
- `libopenscap8` provides libraries enabling integration with SCAP standards
- `bsdmainutils` provides a column command used by usg for list presentation
- `ncal`, dependency of bsdmainutils

jpalermo:
Are these tools that belong on the stemcell? Or are they things that should be used during the stemcell build process to test/validate the stemcells?

The goal of the stemcell is to keep them as lean as possible so we ideally only want packages on there that are needed.

rthill91:
> Are these tools that belong on the stemcell? Or are they things that should be used during the stemcell build process to test/validate the stemcells?
> 
> The goal of the stemcell is to keep them as lean as possible so we ideally only want packages on there that are needed.

There was discussion of providing additional stemcell hardening support (specifically at the time CIS Level 2). Adding this to the build process was considered, but as any changes made after hardening negate it, the cost/benefit just didn't add up. As a partial measure we want to add `usg` to the stemcell so users of the fips stemcell are able to readily run their audits and apply their own relevant tailoring profiles after launch.

aramprice:
Could the same outcome be achieved by installing these tools via a bosh release?

jpalermo:
Yeah, building on @aramprice's point, if this were added to the stemcell, it still can't really be used by end users without some other bosh-release being co-located to actually use the tools.

So if you're going to have to build a bosh-release to use the tools, wouldn't it be cleaner just to have the tools added as part of that bosh release?

rthill91:
> Yeah, building on @aramprice's point, if this were added to the stemcell, it still can't really be used by end users without some other bosh-release being co-located to actually use the tools.
> 
> So if you're going to have to build a bosh-release to use the tools, wouldn't it be cleaner just to have the tools added as part of that bosh release?

by "some other bosh-release" are you referring to the tailoring file that would be used by usg? Assuming so, the tailoring file would be unique to each user/deployment and usg is just the common tool among them. Other concerns/thoughts are
- I'm not sure that we'd be owning/crafting those tailoring files (so not currently able to commit to that effort)
- That release would be an additional place to handle ubuntu pro key management and distribution

rkoster:
What I have seen used before for cases like this is people using the generic [pre-start job from the os-conf release](https://github.com/cloudfoundry/os-conf-release/tree/master/jobs/pre-start-script) installed via a bosh runtime-config to perform these types of operations.

beyhan:
@rthill91 ping ^

rthill91:
Thanks for the feedback and possible alternatives. It seems like having interested partners bake usg into their existing hardening releases is likely the way to go. If anything changes in the future I'd be happy to re-propose this (or a similar) change.

Thanks again for the guidance

PR #329, opened 2024-03-07
Revert from hwe kernel to normal lts kernel

### LINKED ISSUE #318: Memory Errors with stemcell 1.351+
In the past days we saw memory errors with the newer stemcells 1.351+. Let's use this ticket to document all findings and decide how to mitigate this issue.

Slack discussion: https://cloudfoundry.slack.com/archives/C02HWMDUQ/p1707492827160649

jpalermo:
We've seen OOM problems when running cgroups v1 with the 6.5 kernel line.

Resolves #318

The plan is to test this before releasing it. So we'd like to merge it and then use the test stemcells to verify hardware compatibility with the kernel.

Once the stemcells are built in the pipeline, we'll reference the URLs in #318 so people have a chance to take a look at them.

Earliest date we'd want to release this is Monday March 11th.

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy-kernel.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy-kernel.txt
@@ -1,4 +1,4 @@
-bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy-kernel.txtlinux-generic
+linux-generic
 linux-headers-5.15
 linux-headers-5.15-generic
 linux-headers-generic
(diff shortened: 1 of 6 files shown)


lnguyen (dismissed):
:+1: LGTM

rkoster (dismissed):

ramonskie:
you need to adjust https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/stages/static_libraries_config/apply.sh#L16 to 5.15
and this probably needs a check as well https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/lib/prelude_apply.bash#L50

PR #332, opened 2024-03-13
Install and expect ubuntu-pro-client rather than ubuntu-advantage-tools

rthill91:
ubuntu-advantage-tools is being replaced with ubuntu-pro-client. ubuntu-advantage-tools is now a transitional package that just installs ubuntu-pro-client, but will still be installed while in a transitional state.

This likely affects (or will soon affect) bionic, but I haven't tested there

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu-jammy.txt
@@ -458,6 +458,7 @@ tcpdump
 traceroute
 tzdata
 ubuntu-advantage-tools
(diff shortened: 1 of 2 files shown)


ramonskie:
well bionic is not supported anymore for a while now.
but it seems that in jammy they changed the packages
see CI https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/stemcells-ubuntu-jammy/jobs/build-google-kvm-master/builds/24#L65d75897:6089

ramonskie (approved):

rkoster (approved):

PR #334, opened 2024-03-18
Documents why we're now using the non-HWE kernel

klakin-pivotal:
This commit adds some small justification for switching away from the HWE kernel and going back to the -generic kernels, as well as some notes about who to contact if one is considering switching back to the HWE kernel.

This is being added so that someone 6, 12, 24 months from now will have at least SOME reason why we decided to use the LTS kernels over the HWE kernels, and have an idea of who to talk to if they want to revert this decision.


rkoster (approved):

jpalermo (approved):
👍

PR #336, opened 2024-03-22
Create and bundle spdx and cyclonedx json SBOMs as part of stemcell creation

rthill91:
- Install syft in image via Dockerfile
- Add new sbom_create stage to generate sbom.spdx.json as part of stemcell creation
- Bundle sbom.spdx.json into resulting stemcell tgz
- Bundle sbom.cdx.json into resulting stemcell tgz

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb
@@ -71,7 +71,7 @@ module Bosh
         stemcell_name = ArchiveFilename.new(version, definition, 'bosh-stemcell', disk_format).to_s
         tarball_name = File.join(tarball_path, stemcell_name)
 
-        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image']
+        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json']
         Dir.chdir(stemcell_build_path) do
           stdout, stderr, status = Open3.capture3('ls')
           raise stderr unless status.success?
--- a/bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb
@@ -71,7 +71,7 @@ module Bosh
         stemcell_name = ArchiveFilename.new(version, definition, 'bosh-stemcell', disk_format).to_s
         tarball_name = File.join(tarball_path, stemcell_name)
 
-        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json']
+        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json', 'sbom.cdx.json']
         Dir.chdir(stemcell_build_path) do
           stdout, stderr, status = Open3.capture3('ls')
           raise stderr unless status.success?
--- a/ci/docker/os-image-stemcell-builder-jammy/Dockerfile
+++ b/ci/docker/os-image-stemcell-builder-jammy/Dockerfile
@@ -56,6 +56,9 @@ RUN groupadd -o -g ${GROUP_ID} ubuntu                  \
   && useradd -u ${USER_ID} -g ${GROUP_ID} -m ubuntu \
   && echo 'ubuntu ALL=NOPASSWD:ALL' >> /etc/sudoers
 
+# install syft for sbom generation
+RUN curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+
 # we install a local ruby (3.x) because Jammy is lagging at 2.7
 ENV RUBY_SHA256="d2f4577306e6dd932259693233141e5c3ec13622c95b75996541b8d5b68b28b4"
 ENV RUBY_VERSION="3.2.0"
--- a/ci/docker/os-image-stemcell-builder-jammy/Dockerfile
+++ b/ci/docker/os-image-stemcell-builder-jammy/Dockerfile
@@ -57,7 +57,9 @@ RUN groupadd -o -g ${GROUP_ID} ubuntu                  \
   && echo 'ubuntu ALL=NOPASSWD:ALL' >> /etc/sudoers
 
 # install syft for sbom generation
-RUN curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+RUN \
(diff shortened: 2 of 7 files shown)


ci/docker/os-image-stemcell-builder-jammy/Dockerfile
```diff
@@ -56,6 +56,9 @@ RUN groupadd -o -g ${GROUP_ID} ubuntu                  \
   && useradd -u ${USER_ID} -g ${GROUP_ID} -m ubuntu \
   && echo 'ubuntu ALL=NOPASSWD:ALL' >> /etc/sudoers
 
+# install syft for sbom generation
+RUN curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
```
rkoster: Please use an immutable URL and digest verification.

rkoster (changes requested):
PR looks great! Not sure if we want to do some filtering to reduce false positives (at some point the google agent tools had a dev dependency which would be picked up by syft). But those could also be done at a later point in time, once people start consuming the sbom and provide feedback.

ragaskar:
I agree w/ @rkoster that there should be some provision for checking the provenance of the install script/installed libraries. 

I can also share that in our experience, we've seen issues with tracking latest `syft` in that new features are added very quickly and occasionally these new features have undesirable impacts (e.g., we've seen new cataloguers suddenly introduce records for dozens of libraries that -- in our case -- were all false positives). It might be nice to have a way to easily pin syft when necessary. It's possible there is some implementation overlap with provenance for version pinning, so perhaps it makes sense to take that on now.

ragaskar (dismissed):
See my comment above, I see @rkoster is already asking for some validation strategy with the syft install, which I agree with. 

I don't think the version-pinning provision I mentioned is required so I'll give this a thumbs up, but it's a definite nice-to-have that may intersect with the provenance piece.

ragaskar:
also, thanks for this contribution, cool stuff!

ci/docker/os-image-stemcell-builder-jammy/Dockerfile
```diff
@@ -56,6 +56,15 @@ RUN groupadd -o -g ${GROUP_ID} ubuntu                  \
...
   && echo 'ubuntu ALL=NOPASSWD:ALL' >> /etc/sudoers
 
+# install syft for sbom generation
+ADD syft.checksum /tmp/syft.checksum
+ADD https://raw.githubusercontent.com/anchore/syft/5a865d0d904b90252c02e0cceeac61c3d882837d/install.sh /tmp/install.sh
+RUN \
+    cd tmp \
+    sha256sum -c --status syft.checksum && \
+    sh install.sh -b /usr/local/bin && \
+    rm install.sh syft.checksum
```
rkoster: I would prefer a simple one-liner like: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/4481987e71b816ae0bf7f20119adc3aa7f610cf4/ci/docker/os-image-stemcell-builder-jammy/Dockerfile#L113-L116 over downloading and executing an external install script.

beyhan:
What is the estimated size of the generated SBOM? Have you also tried the other formats supported like the cycloneDX and the syft own format?

rthill91:
> What is the estimated size of the generated SBOM? Have you also tried the other formats supported like the cycloneDX and the syft own format?

~34M in my jammy fips test build from today.

As for other formats, minimal consideration was given beyond the desire for something generic and machine readable. That put syft format out. Beyond that, internally spdx is the chosen format.

rkoster (dismissed):

VinodAnandan:
@rthill91 Could you please consider adding support for CycloneDX format as well?

rkoster:
Our internal tooling uses CycloneDX, so supporting both formats would be a plus for us.
The `cyclonedx-json` output format [seems to be supported by syft](https://github.com/anchore/syft?tab=readme-ov-file#output-formats).
But this is not a blocker for me and could be done by someone else in a later PR. I just want to make sure the discussion about supporting multiple formats is captured here.

rthill91:
> Our internal tooling uses CycloneDX, so supporting both formats would be a plus for us. The `cyclonedx-json` output format [seems to be supported by syft](https://github.com/anchore/syft?tab=readme-ov-file#output-formats). But this is not a blocker for me and could be done by someone else in a later PR. I just want to make sure the discussion about supporting multiple formats is captured here.

syft definitely supports cyclonedx. The only complexity (off the top of my head) to adding support is that it makes it turns sbom generation into a configurable option rather than a static thing and it'll make the expected files in the output tgz configurable. I could just change the sbom to `sbom.json` to avoid that last bit, but I like the clarity of specifying that it's spdx, particularly if there's support for other formats included.

I don't mind working this, but it'll likely have to wait a few weeks (Noble release is going to keep me busy) and I'd rather it be a separate PR.

rkoster:
I would not make it an optional build flag, but instead just include both sboms, similar to what buildpacks are doing: https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md#what-it-is

So we would end up with something like:
```ruby
expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json', 'sbom.cdx.json']
```

rthill91:
> I would not make it an optional build flag, but instead just include both sboms, similar to what buildpacks are doing: https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md#what-it-is
> 
> So we would end up with something like:
> 
> ```ruby
> expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json', 'sbom.cdx.json']
> ```

including both makes this a trivial change. I have it staged locally but I'm seeing some test failures (going through the non-iaas-kernel tests rather than aws specific kernel). I'll have to look at this a bit later

rthill91:
updated to generate cyclonedx and spdx sboms

bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb
```diff
@@ -71,7 +71,7 @@ def create_tarball(disk_format)
         stemcell_name = ArchiveFilename.new(version, definition, 'bosh-stemcell', disk_format).to_s
         tarball_name = File.join(tarball_path, stemcell_name)
 
-        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image']
+        expected = ['stemcell.MF', 'packages.txt', 'dev_tools_file_list.txt', 'image', 'sbom.spdx.json', 'sbom.cyclonedx.json']
```
rkoster: According to: https://cyclonedx.org/specification/overview/#recognized-file-patterns
the filename should match `*.cdx.json`

rkoster (changes requested):

rkoster (approved):

rthill91:
updated title/description to reflect the addition of cdx sbom

ragaskar (approved):
thank you, looks great.

rkoster:
Thanks! @rthill91

PR #348, opened 2024-05-14
Add EFI support for vsphere stemcell

lnguyen:
This is for vSphere stemcells can to access > 16 GiB GPUs

Booting from an EFI partition is necessary to access and utilize
GPUs with more than 16 GB of memory because UEFI provides the
advanced memory addressing, compatibility, and initialization
capabilities required for such hardware. The older BIOS/MBR
system cannot efficiently handle the requirements of modern
high-capacity GPUs.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/disk_image.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/disk_image.rb
@@ -14,17 +14,16 @@ module Bosh::Stemcell
     end
 
     def mount
-      device_path   = stemcell_loopback_device_name
-      mount_command = "sudo mount #{device_path} #{image_mount_point}"
-      shell.run(mount_command, output_command: verbose)
+      mount_image
     rescue => e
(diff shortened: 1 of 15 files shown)


jpalermo (approved):
👍

cunnie:
I love this commit, but could you please modify the message (which does a great job of explaining what you did, but glosses over _why_ you did it):

> vSphere stemcells can access > 16 GiB GPUs

> booting from an EFI partition is necessary to access and utilize GPUs with more than 16 GB of memory because UEFI provides the advanced memory addressing, compatibility, and initialization capabilities required for such hardware. The older BIOS/MBR system cannot efficiently handle the requirements of modern high-capacity GPUs.

cunnie (changes requested):
Could you please update your commit message? One that explains _why_ this change was necessary.

cunnie (approved):
Thanks for updating the commit message! Looks good.

PR #349, opened 2024-05-15
Upload sboms to github

rkoster:
Based on example from here: https://docs.github.com/en/enterprise-cloud@latest/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api


jpalermo (approved):
👍

PR #356, opened 2024-06-04
retries for firstboot and let ssh.service depends on firstboot_done file

### LINKED ISSUE #337: Risk of false positive on systemd rc.local logics when execution of firstboot.sh fails
Existing implementation on systemd [rc.local](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/master/stemcell_builder/stages/base_ubuntu_firstboot/assets/etc/rc.local#L5) could have the risk of false positive if execution of `firstboot.sh` fails. This is because execution of `firstboot.sh` is in a new shell and it does not inherit the `set -e` option from `rc.local`. The `rc.local` still exits with status `0` even when execution of `firstboot.sh` fails. When it happens, there is no host key generated that then result in ssh failure on the provisioned VM.

Would it be feasible to add a retry logic on `firstboot.sh` execution failure in `rc.local` to mitigate the potential risk of false positive?

ramonskie:
this will fix #337

### DIFF
--- a/stemcell_builder/stages/base_ssh/apply.sh
+++ b/stemcell_builder/stages/base_ssh/apply.sh
@@ -100,3 +100,7 @@ then
 else
   sudo echo "ENABLED=0" >$chroot/etc/default/motd-news
 fi
+
+# add drop-in for sshd service so it will only starts after first boot has been succefully run
+mkdir -p $chroot/lib/systemd/system/ssh.service.d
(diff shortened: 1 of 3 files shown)


aramprice (approved):

ramonskie:
@mvach @klakin-pivotal  could either of you please review this

jpalermo (approved):
👍

ystros:
FYI @ramonskie @jpalermo , we've reverted this change because it was causing BATS failures in the `jammy-master` sub-pipeline. This was blocking other bumps from being promoted to the 1.x branch. See #366 for more details.

PR #357, opened 2024-06-05
Updates validation test for new GRUB conf location

klakin-pivotal:
This commit updates the GRUB bootloader location and contents verification test for the FIPS stemcell creator to look for the file in a new location.

This change to the location of the GRUB config was probably introduced in commit b67cf289f84757bc708be45aeb1235462ef7b373.

This commit attempts to correct the following error in CI:

=====
```
Failures:

  1) FIPS Stemcell installed by image_install_grub for fips kernel /boot/grub/grub.cfg is expected to be file
     Failure/Error: it { should be_file }
       expected `#<ShelloutTypes::File:0x00007fd51bdd6718 @path="/boot/grub/grub.cfg", @chroot=#<ShelloutTypes::Chroot:0x00007fd51bdd6740>>.file?` to be truthy, got false
     # ./spec/stemcells/fips_spec.rb:40:in `block (4 levels) in <top (required)>'

  2) FIPS Stemcell installed by image_install_grub for fips kernel /boot/grub/grub.cfg content
     Failure/Error: raise RuntimeError, stderr if status != 0

     RuntimeError:
       cat: /boot/grub/grub.cfg: No such file or directory
     # ./lib/shellout_types/file.rb:42:in `content'
     # ./spec/stemcells/fips_spec.rb:41:in `block (4 levels) in <top (required)>'

  3) FIPS Stemcell installed by image_install_grub for fips kernel /boot/grub/grub.cfg content
     Failure/Error: raise RuntimeError, stderr if status != 0

     RuntimeError:
       cat: /boot/grub/grub.cfg: No such file or directory
     # ./lib/shellout_types/file.rb:42:in `content'
     # ./spec/stemcells/fips_spec.rb:45:in `block (4 levels) in <top (required)>'
```

PR #365, opened 2024-07-10
fix: determine grub.cfg location based on infrastructure

rthill91:
The change in location of grub.cfg only impacts vsphere however the test change (https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/357) made no distinction. This commit uses the infrastructure to determine which location should be checked in test.


rkoster (approved):

rkoster:
Thanks! @rthill91

PR #366, opened 2024-07-18
Revert "retries for firstboot and let ssh.service depends on firstboot_done file"

ystros:
Reverts cloudfoundry/bosh-linux-stemcell-builder#356

We are reverting this because it appears to cause failures when running BATS. BATS is unable to use `bosh ssh` to SSH into the deployed VM. We tried replicating this by manually downloading and deploying the stemcell, but were unable to reproduce the failure.

See this failed build for more details:
https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/stemcells-ubuntu-jammy/jobs/bats-master/builds/147

### DIFF
--- a/stemcell_builder/stages/base_ssh/apply.sh
+++ b/stemcell_builder/stages/base_ssh/apply.sh
@@ -100,7 +100,3 @@ then
 else
   sudo echo "ENABLED=0" >$chroot/etc/default/motd-news
 fi
-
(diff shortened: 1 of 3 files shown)


selzoc (approved):

klakin-pivotal (approved):
Given that we're reverting this to see if it gets rid of a persistent failure in CI, LGTM.

PR #372, opened 2024-07-24
Remove non-jammy code / conditionals from the branch we use to create Jammy stemcells

aramprice:
The `ubuntu-jammy/...` branches are not used to build stemcells of other unbutu versions so there is no need to carry around this cruft.


lnguyen (dismissed):

ramonskie:
Yeah i also removed it from the noble branch. Can you double check there as well if we did the same?

aramprice:
@ramonskie is that branch `ubuntu_noble_poc`?
> Q: if so could we rename it to better match the other branches?
> 

Something like `ubuntu-noble/main` in keeping with the ongoing transition from `master` => `main`

ramonskie:
Yes currently its ubunt_noble_poc.
I did this as if we put it under ubunt-noble/master rules will be enforced like reviewers are needed etc. Once we have a working noble stemcell we can put it under the Pröpper branch structure

aramprice:
@ramonskie - probably the easiest will be to rebase from `ubuntu-jammy/master` after this change lands, or cherry-pick these changes.

jpalermo (approved):
Much cleaner. Thanks!

PR #374, opened 2024-07-26
Lower ClientAliveInterval for ssh

ystros:
The Azure Marketplace requires that images have set ClientAliveInterval to between 30 and 235, with a recommended value of 180. It seems safe to update the value for all IaaSes, as it simply controls how long to wait before the sshd server sends a message to a quiet client asking for a response.

For more information, see:
https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-marketplace-images#linux-and-open-source-os-images

### DIFF
--- a/stemcell_builder/stages/system_azure_init/assets/etc/waagent/waagent.conf
+++ b/stemcell_builder/stages/system_azure_init/assets/etc/waagent/waagent.conf
@@ -85,8 +85,8 @@ OS.OpensslPath=None
 # Set the path to SSH keys and configuration files
 OS.SshDir=/etc/ssh
 
-# Set /etc/ssh/sshd_config ClientAliveInterval to 900
-OS.SshClientAliveInterval=900 # NON-DEFAULT, works around azure TCP NAT tables timeouts
+# Set /etc/ssh/sshd_config ClientAliveInterval to 180
+OS.SshClientAliveInterval=180
 
 # If set, agent will use proxy server to access internet
 #HttpProxy.Host=None
--- a/bosh-stemcell/spec/support/os_image_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_shared_examples.rb
@@ -279,8 +279,8 @@ shared_examples_for 'every OS image' do
       expect(sshd_config.content).to match(/^IgnoreRhosts yes$/)
     end
 
-    it 'sets ClientAliveInterval to 900 seconds (CIS-5.2.12)' do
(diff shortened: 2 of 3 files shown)


stemcell_builder/stages/system_azure_init/assets/etc/waagent/waagent.conf
```diff
@@ -85,8 +85,8 @@ OS.OpensslPath=None
 # Set the path to SSH keys and configuration files
 OS.SshDir=/etc/ssh
 
-# Set /etc/ssh/sshd_config ClientAliveInterval to 900
-OS.SshClientAliveInterval=900 # NON-DEFAULT, works around azure TCP NAT tables timeouts
```
selzoc, line 89: Are we concerned about the comment here about `TCP NAT tables timeouts` ?

ystros: This is a good question. I *think* we will be okay, but it is not 100% clear. It is difficult to find context on this "TCP NAT table timeout" issue referenced in the comment.

The interval was originally set in `waagent.conf` as part of https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/88, which removed ClientAliveInterval from the `sshd_config`. However, it was added back to the `sshd_config` soon after in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/5d8f198eb23d1a83e010e8f4dbc06d9428b5619b and later was lowered only in the `sshd_config` as part of https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/111

I'm unsure of how the settings in `waagent.conf` get applied, but can definitely confirm that the contents of `sshd_config` on a BOSH-deployed Azure VM match those generated by `base_ssh/apply.sh`, and don't seem to be overwritten by the waagent settings. Which implies we've been running with an interval of 300 second since March 2020 with no reports of problems. Since we don't know what the original problem was, it's unclear if the problem is just gone now or if 180 will be "too low" and cause the problem to recur.

```
test-bu-1/83f80230-0db9-40ab-b588-d14161ea97b3:~# cat /etc/ssh/sshd_config

# This is the sshd server system-wide configuration file.  See
# sshd_config(5) for more information.

... snip ...

IgnoreRhosts yes
ClientAliveInterval 300
LoginGraceTime 60

... snip ...
```

selzoc (approved):

aramprice (approved):

PR #384, opened 2024-10-25
Restore default apt sources once stemcell is built

selzoc:
We use the snapshot.ubuntu.com sources for build repeatability, but for products that are built on top of stemcells, we want to be able to get the latest updates to packages.

This adds a stage called "restore_apt_sources" that runs after the stemcell is constructed.


jpalermo (approved):
👍

aramprice (approved):

ramonskie:
personally. i thing updating apt and packages for updates on a stemcell is a no go.
but i understand the reasoning behind it that sometimes its just needed for a quickfix.

ramonskie (approved):

PR #390, opened 2024-11-15
Noble fallbackbootloader

ramonskie:
this will add a legacy fallback bootloader to the efi bootloader.

we deleted the old image_install_grub and renamed the image_install_grub_efi to image_install_grub
so we can keep just one configuration stage for all iaas stemcells.

we needed to have this fallback. as currently some older vm types in aws and probably other iaas's as well
that currently do not support efi bootloader


benjaminguttmann-avtq (approved):
lgtm

lnguyen (approved):

jpalermo:
I feel strongly that I have no idea what is going on here... but if it works, it works 🤷

jpalermo (approved):

PR #387, opened 2024-11-06
change pam_lastlog for pam_faillock + remove adiscon rsyslog ppa

ramonskie:
this will change the deprecated pam_lastlog which logged showfailed logins
to pam_faillock which also shows failed logins in a similar matter.
this will solve #343 

this pr also removed the rsyslog adiscon ppa. as we now use the rsyslog provided in the main ubuntu package repo.
this also contributes to better reproduciblty 
this wil solve #299

PR #393, opened 2024-11-22
Revert "change pam_lastlog for pam_faillock + remove adiscon rsyslog ppa"

ramonskie:
Reverts cloudfoundry/bosh-linux-stemcell-builder#387

this breaks now as a user is not able to login in anymore


rkoster (approved):

ramonskie:
we can implement something like this
```
session required pam_faillock.so preauth audit silent deny=3 unlock_time=600
session required pam_faillock.so authfail audit deny=3 unlock_time=600
```
this will and log and lockout the account. the logs are a bit different though.

jpalermo:
[Original addition](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/1468cd27059434639683272bc90fad4ae7f017f4) seems to have been stig related, so if we don’t add this back some security scanners are going to get cranky

jpalermo (approved):

PR #395, opened 2024-11-25
Noble remove centos opensuse

ramonskie:
this will remove alle CentOS en OpenSuse specific build stuff.

please  note the commit about remove su root command for syslog.
it seemed that we implemented this for centos specific but used it for all builds.
i think its safe to say we can remove this. and revert back to default behavior.
but this could potentiality break something


benjaminguttmann-avtq (commented):
found one more appearance of centos within the Vagrantfile (https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/f3746b30724ee3c499168cdcec9bf37f549af382/ci/docker/os-image-stemcell-builder/Vagrantfile#L48) but at least in the ubuntu-jammy branch this does not exist anymore.

ramonskie:
yeah i think we can get rid of vagrant as well as nobody uses it for a very long time

aramprice (approved):

benjaminguttmann-avtq (approved):

jpalermo (approved):

PR #397, opened 2024-11-26
Update warden stage to support Noble

jpalermo:
- Enable systemd in the container
- Remove all runit shims that were previously included because systemd was not being used
- local_events must be disabled for auditd due to running in a container
- systemd DefaultStartLimitBurst must be increased because some services are restarted more than 5 times in 5 seconds on agent startup (systemd-resolvd was seen at least)
- Disable pam_faillock module. This was a differnet pam module in Jammy and prior. It's used to show number of previous failed login attempts. It seems that when it's enabled within a container sshd is unable to properly tracker users and causes ssh-ing into the container to fail.


ystros (dismissed):

aramprice (dismissed):

ramonskie (dismissed):
tested it and work

beyhan (approved):

ramonskie (approved):
lgtm

PR #398, opened 2024-12-04
set pam_lastlog2 optional

ramonskie:
if util-linux 2.40 will be backported to noble. or the user provide the pam_lastlog2.so them self this setting wil then be used

### DIFF
--- a/bosh-stemcell/spec/os_image/ubuntu_noble_spec.rb
+++ b/bosh-stemcell/spec/os_image/ubuntu_noble_spec.rb
@@ -292,7 +292,7 @@ EOF
 
   context 'display the number of unsuccessful logon/access attempts since the last successful logon/access (stig: V-51875)' do
     describe file('/etc/pam.d/common-password') do
-      its(:content) { should match /session\trequired\t\t\tpam_lastlog\.so showfailed/ }
+      its(:content) { should match /session\toptional\t\t\tpam_lastlog2\.so showfailed/ }
     end
   end
 
--- a/stemcell_builder/stages/password_policies/assets/ubuntu/common-password.patch
+++ b/stemcell_builder/stages/password_policies/assets/ubuntu/common-password.patch
@@ -13,5 +13,5 @@
  # since the modules above will each just jump around
  password	required			pam_permit.so
  # and here are more per-package modules (the "Additional" block)
-+# session	required			pam_lastlog.so showfailed #NOBLE_TODO: Commented out when pam_lastlog is availble again
++session	optional			pam_lastlog2.so showfailed #NOBLE_TODO: this will only work if util-linux =>2.40 which provide pam_lastlog2.so or if users will install it manually
  # end of pam-auth-update config


beyhan (approved):

jpalermo (approved):

aramprice:
This may be causing some specs to fail[here](https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/stemcells-ubuntu-noble/jobs/build-warden-boshlite/builds/55#L673c6a85:14995:15045)

```
  1) Warden Stemcell pam common-password config /etc/pam.d/common-password content is expected to include "#session\trequired\t\t\tpam_faillock.so"
     Failure/Error: its(:content) { should include('#session	required			pam_faillock.so') }
     
       expected "#\n# /etc/pam.d/common-password - password-related modules common to all services\n#\n# This file is included from other service-specific PAM config files,\n# and should contain a list of modules that define the services to be\n# used to change user passwords.  The default is pam_unix.\n\n# Explanation of pam_unix options:\n# The \"yescrypt\" option enables\n#hashed passwords using the yescrypt algorithm, introduced in Debian\n#11.  Without this option, the default is Unix crypt.  Prior releases\n#used the option \"sha512\"; if a shadow password hash will be shared\n#between Debian 11 and older releases replace \"yescrypt\" with \"sha512\"\n#for compatibility .  The \"obscure\" option replaces the old\n#`OBSCURE_CHECKS_ENAB' option in login.defs.  See the pam_unix manpage\n#for other options.\n\n# As of pam 1.0.1-6, this file is managed by pam-auth-update by default.\n# To take advantage of this, it is recommended that you configure any\n# local modules either before or after the default block, and use\n# pam-auth-update to manage selection of other modules.  See\n# pam-auth-update(8) for details.\n\n# here are the per-package modules (the \"Primary\" block)\npassword\trequisite\t\t\tpam_pwquality.so retry=3 minlen=14 dcredit=-1 ucredit=-1 ocredit=-1 lcredit=-1 difok=8\npassword\t[success=1 default=ignore]\tpam_unix.so obscure use_authtok try_first_pass sha512 remember=24 minlen=14 rounds=5000\n# here's the fallback if no module succeeds\npassword\trequisite\t\t\tpam_deny.so\n# prime the stack with a positive return value if there isn't one already;\n# this avoids us returning an error just because nothing sets a success code\n# since the modules above will each just jump around\npassword\trequired\t\t\tpam_permit.so\n# and here are more per-package modules (the \"Additional\" block)\nsession\toptional\t\t\tpam_lastlog2.so showfailed #NOBLE_TODO: this will only work if util-linux =>2.40 which provide pam_lastlog2.so or if users will install it manually\n# end of pam-auth-update config\n" to include "#session\trequired\t\t\tpam_faillock.so"
       Diff:
       @@ -1,35 +1,69 @@
       -#session	required			pam_faillock.so
       +#
       +# /etc/pam.d/common-password - password-related modules common to all services
       +#
       +# This file is included from other service-specific PAM config files,
       +# and should contain a list of modules that define the services to be
       +# used to change user passwords.  The default is pam_unix.
       +
       +# Explanation of pam_unix options:
       +# The "yescrypt" option enables
       +#hashed passwords using the yescrypt algorithm, introduced in Debian
       +#11.  Without this option, the default is Unix crypt.  Prior releases
       +#used the option "sha512"; if a shadow password hash will be shared
       +#between Debian 11 and older releases replace "yescrypt" with "sha512"
       +#for compatibility .  The "obscure" option replaces the old
       +#`OBSCURE_CHECKS_ENAB' option in login.defs.  See the pam_unix manpage
       +#for other options.
       +
       +# As of pam 1.0.1-6, this file is managed by pam-auth-update by default.
       +# To take advantage of this, it is recommended that you configure any
       +# local modules either before or after the default block, and use
       +# pam-auth-update to manage selection of other modules.  See
       +# pam-auth-update(8) for details.
       +
       +# here are the per-package modules (the "Primary" block)
       +password	requisite			pam_pwquality.so retry=3 minlen=14 dcredit=-1 ucredit=-1 ocredit=-1 lcredit=-1 difok=8
       +password	[success=1 default=ignore]	pam_unix.so obscure use_authtok try_first_pass sha512 remember=24 minlen=14 rounds=5000
       +# here's the fallback if no module succeeds
       +password	requisite			pam_deny.so
       +# prime the stack with a positive return value if there isn't one already;
       +# this avoids us returning an error just because nothing sets a success code
       +# since the modules above will each just jump around
       +password	required			pam_permit.so
       +# and here are more per-package modules (the "Additional" block)
       +session	optional			pam_lastlog2.so showfailed #NOBLE_TODO: this will only work if util-linux =>2.40 which provide pam_lastlog2.so or if users will install it manually
       +# end of pam-auth-update config
       
     # ./spec/stemcells/warden_spec.rb:26:in `block (4 levels) in <top (required)>'

Finished in 45.75 seconds (files took 3.65 seconds to load)
505 examples, 1 failure, 1 pending

Failed examples:

rspec ./spec/stemcells/warden_spec.rb:26 # Warden Stemcell pam common-password config /etc/pam.d/common-password content is expected to include "#session\trequired\t\t\tpam_faillock.so"
```

ramonskie:
yeah there where some change when we merged the warden fix.
so we will need address this again for the warden.spec

mingxiao:
> so we will need address this again for the warden.spec

Are you planning to address that? @ramonskie

ramonskie:
@mingxiao i can pick it up tomorrow. But if you have time you could also make a pr

mingxiao:
> i can pick it up tomorrow.

sounds good. We aren't sure what changes needs to be made

ramonskie:
@mingxiao fixed in #399

ramonskie:
currently our tests are failing due to the fact that if the pam_lastlog2.pam is not there it will be a noisy output in the auth.log when logging in
```
2024-12-19T13:30:01.348861+00:00 0108b3e2-aa96-4287-b7f3-4a4e0dddb200 CRON[5798]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
2024-12-19T13:30:01.353994+00:00 0108b3e2-aa96-4287-b7f3-4a4e0dddb200 CRON[5798]: pam_unix(cron:session): session closed for user root
2024-12-19T13:35:01.360616+00:00 0108b3e2-aa96-4287-b7f3-4a4e0dddb200 CRON[5828]: PAM unable to dlopen(pam_lastlog2.so): /usr/lib/security/pam_lastlog2.so: cannot open shared object file: No such file or directory
```
and our tests check for the log files if it contains `No such file or directory`

the only suggestion i have is to disable the pam_lastlog2 module again. is its not available anyway.
and for people who do want lastlog information they could retrieve this from systemd `journalctl _COMM=sshd`

PR #400, opened 2024-12-12
remove partitionar_type_mapping as we dont use it 

ramonskie:
for the past 5 year we have not used this function that coul switch between partitionar type.
and as parted is now the only partitioner program we used. so by removing this we make the code a little bit less complex

### DIFF
--- a/stemcell_builder/stages/bosh_cloudstack_agent_settings/apply.sh
+++ b/stemcell_builder/stages/bosh_cloudstack_agent_settings/apply.sh
@@ -2,7 +2,6 @@
 
 base_dir=$(readlink -nf $(dirname $0)/../..)
 source $base_dir/lib/prelude_apply.bash
-source $base_dir/lib/prelude_agent.bash
 
 agent_settings_file=$chroot/var/vcap/bosh/agent.json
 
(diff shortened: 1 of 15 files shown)


aramprice (dismissed):

ramonskie:
as we remove this. would it not also be wise to put the agent.json in a assets folder and copy it from there.
it thing this will make it allot more visible. and remove some complexity with heardocs.
why i bring this up as we are currently investigating putting regex values in the agent.json
for better disk path resolver per iaas.
and currently you need to comment out multiple `${1}`  which should ofcourse not be interpolated

stemcell_builder/stages/bosh_cloudstack_agent_settings/apply.sh
```diff
@@ -2,15 +2,14 @@
...
-source $base_dir/lib/prelude_agent.bash
 
 agent_settings_file=$chroot/var/vcap/bosh/agent.json
 
 cat > $agent_settings_file <<JSON
 {
   "Platform": {
     "Linux": {
-      $(get_partitioner_type_mapping)
+      "PartitionerType": "parted",
```
aramprice: Could this also be moved in to an assets file? It doesn't look like there's any more interpolation in the `<<JSON` heredoc.

aramprice (dismissed):

rkoster (approved):

jpalermo (approved):

PR #401, opened 2024-12-12
refactor stemcell repack script

ramonskie:
this script can inject a new `bosh-agent`, `agent json` or add a public debug ssh key to connect to the vm if the agent does not start


aramprice (approved):

scripts/stemcell-repack-debuger.sh
```diff
@@ -0,0 +1,125 @@
...
+done
+
+if [ -n "$convert_cert_file" ]; then
+    BOSH_DEBUG_PUB_KEY=$(convert_multiline_to_single "$convert_cert_file")
+    export BOSH_DEBUG_PUB_KEY
+fi
+
+
+if [ -z ${stemcell_tgz} ]; then
+echo "stemcell_tgz is not set. export stemcell_tgz="/home/username/workspace/bosh/bosh-linux-stemcell-builder/tmp/bosh-stemcell-0.0.8-google-kvm-ubuntu-noble-go_agent.tgz""
```
benjaminguttmann-avtq, line 68: does it make sense that the version is hardcoded here?

ramonskie: the version number corresponds with what we use in the readme for building stemcells
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/README.md?plain=1#L67

benjaminguttmann-avtq (approved):

lnguyen (approved):

PR #407, opened 2025-01-15
Update 60-bosh-sysctl.conf

fmoehler:
Our lynis scans reported this as a potential vulnerability. We tested that the ipv6 connection still works after this is disabled.

### DIFF
--- a/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf
+++ b/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf
@@ -22,8 +22,8 @@ net.ipv4.icmp_ignore_bogus_error_responses=1
 
 net.ipv6.conf.all.disable_ipv6=0
 net.ipv6.conf.default.disable_ipv6=0
-net.ipv6.conf.default.accept_redirects=1
-net.ipv6.conf.all.accept_redirects=1
+net.ipv6.conf.default.accept_redirects=0
+net.ipv6.conf.all.accept_redirects=0
 net.ipv6.route.flush=0
 
 kernel.exec-shield=1


beyhan:
This needs further checks and clarifications @xtreme-nitin-ravindran would be great if you could also check from security perspective the change

fmoehler:
Just as a note: This setting was disabled by default in jammy as well: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/17f05f8c54fd89d8f64d1edb2a050612e5455793/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf#L25

ramonskie:
the idea for noble is that we do not set any ipv6 kernel stuff anymore from the agent.
as it does some weird reboot stuff.

so we want to make ipv6 by default enabled (as is in the upstream ubuntu by default, and we set a kernel flag to disable it in jammy currently)

fmoehler:
Yes but ipv6 will remain [enabled](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/bbce6c2df5568c8b2ca4c62c09cf4757378acde6/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf#L23) in noble, I just want to disable this accept_redirects

fmoehler:
According to this [website](https://docs.datadoghq.com/security/default_rules/xccdf-org-ssgproject-content-rule-sysctl-net-ipv6-conf-all-accept-redirects/) it could be a security risk to have it enabled.

xtreme-nitin-ravindran:
CIS Ubuntu Noble v1.0.0 requires both those kernel parameters along with their IPv4 equivalent to be set to `0`.  All 4 kernel parameters are set to `0` in Jammy.

Here's the CIS text for Noble

**3.3.5 Ensure icmp redirects are not accepted (Automated)**
***Description:***
ICMP redirect messages are packets that convey routing information and tell your host (acting as a router) to send packets via an alternate path. It is a way of allowing an outside routing device to update your system routing tables.

***Rationale:***
ICMP redirect messages are packets that convey routing information and tell your host (acting as a router) to send packets via an alternate path. It is a way of allowing an outside routing device to update your system routing tables. By setting `net.ipv4.conf.all.accept_redirects`, `net.ipv4.conf.default.accept_redirects`, `net.ipv6.conf.all.accept_redirects`, and `net.ipv6.conf.default.accept_redirects` to `0`, the system will not accept any ICMP redirect messages, and therefore, won't allow outsiders to update the system's routing tables.

beyhan:
@xtreme-nitin-ravindran my understanding from your comment is that you agree with this change? For IPv4 the values are already configured in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf#L9 and https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf#L1.

Looking into [60-bosh-sysctl.conf](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf) there are many other options configured for IPv4. Are those not relevant for IPv6? @fmoehler have you checked on those?

peanball:
I read up on `accept_redirects` a bit, having worried that it might be needed for containerized workloads (i.e. Diego). But that is not the case.

Also clearly confirmed by those settings being off in Jammy.

CF does not run L3 routing on the BOSH VMs as far as I understand. And even so, `accept_redirects` is a tweak, not a necessity for routing.

fmoehler:
@beyhan no I have not. The accept_redirect was the only option reported by our lynis scans for the noble stemcell. So I think the rest should be fine.

beyhan:
@fmoehler, @peanball as we want to support IPv6 with Noble I think it make sense to check the configurations available in [60-bosh-sysctl.conf](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-jammy/stemcell_builder/stages/bosh_sysctl/assets/60-bosh-sysctl.conf) and see whether we are missing something. The Noble stemcell is approaching GA and after the GA release changes in the stemcell aren't so easy. I don't say that it should happen in the scope of this issue.

xtreme-nitin-ravindran:
@beyhan: Sorry if I wasn't explicit. I agree with the change, and we need the change to the IPv4 params as well.

fmoehler:
@xtreme-nitin-ravindran that is already set https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/407/files#diff-bbbdb33f10c226d80d7df4f60dbfe667cc962483dec6b250736361953f87bd0eL1

ramonskie (approved):

jpalermo (approved):

PR #411, opened 2025-01-24
Stop using ubuntu release name in filenames

aramprice:
Now that different stemcell releases are on separate branches we do not need to differentiate filenames by os release name.

This commit also removes a number of non Ubuntu Jammy files, there are likely more that could be cleaned up, notably some comments / conditionals in the following files:

- stemcell_builder/stages/system_kernel/apply.sh
- stemcell_builder/stages/rsyslog_config/assets/rsyslog_logrotate.conf

The contents of `bosh-stemcell/image-metalinks/#ubuntu-jammy/` should probably also be examined to see if they can be deleted and/or consumed from some other branch of this repo.


ramonskie:
this change will interfere with our jammy/noble upstream merge
and in the future noble/xxxx

we merge these branches sometime to keep up with the changes that has been made in jammy

benjaminguttmann-avtq (approved):
lgtm

jpalermo:
This will certainly make the next merge between jammy and noble harder, but once that's done it puts us in a position where it will be much easier to merge and could possibly be done automatically at some point.

ramonskie (approved):
approved as discussed in the foundational meeting

fmoehler (approved):

PR #412, opened 2025-01-25
Move the stemcell acceptance specs to top-level

aramprice:
Prior to the go module dependency management tooling Golang source code needed to be in a repository-relative path under `src/`, this is no longer true. This commit moves `src/github.com/cloudfoundry/stemcell-acceptance-tests/` to the top-level of the repo, renaming the folder to `acceptance-tests/`. The golang packages are updated to reflect this.

In addition the `tools` package is removed since Ginkgo v2 no longer requires this to ensure the CLI code is available when testing.


aramprice:
The following should also be merged:
- https://github.com/cloudfoundry/bosh-stemcells-ci/pull/31
- https://github.com/cloudfoundry/bosh-cpi-certification/pull/20

aramprice:
Searched for consumers via the following query
- https://github.com/search?q=org%3Acloudfoundry%20github.com%2Fcloudfoundry%2Fstemcell-acceptance-tests

ramonskie:
looks good to me

as stated in #411 if we are going that route
we need an additional pr for noble

ramonskie (approved):

jpalermo:
We can merge this forward to the Noble branch after it's been merged that way we avoid a second PR

jpalermo (approved):

PR #418, opened 2025-02-28
Decreases "recordfail" timeout to five seconds

klakin-pivotal:
This commit reduces the boot timeout for the "recordfail" GRUB boot logic from its default (30 seconds) to five seconds. The ordinary boot wait time is five seconds, so this makes the boot wait the same regardless of which branch is taken.

Currently, we strongly believe that the first boot of any VM based on a stemcell falls into the "recordfail" branch, and has its startup delayed by 30 seconds for no good reason.

Do note that this environment variable is not documented in the official GRUB documentation, but -then again- most of the environment variables used by the GRUB configuration generation machinery are also undocumented... source code inspection is the only way to learn of their existence.


aramprice:
It might be helpful to include a pointer to the Grub source file/path/something in the commit to aid in future sleuthing.

aramprice (approved):

klakin-pivotal:
Oh, are you asking for a mention of the file that contains the reference to the environment variable(s) that are useful but undocumented? If so, I'll edit the commit message with that information.

Edit: How does this new commit message look? <https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/418/commits/b4a54127f3db73900292a87047cb9f0c246f3d1f>

ramonskie (approved):
looks good.
we should also update this for noble

fmoehler (approved):

Issue #415, opened 2025-02-06
Noble: consider removing scsitools?


anshrupani:
Hi everyone!

We have been continuously maintaining a hardening release for our environments which cleans up packages which are not really important during the lifecycle of a VM, after it starts successfully. One such package is **scsitools**, which is also not a part of the main repository (but the universe repo).

We were wondering, that if this really might not be necessary, should we consider removing scsitools from the noble stemcell?

Since I am also not very well versed with the usecase for this package, it would be great if you guys would have some input regarding the same.

Follow up on #328

a-hassanin:
Hi @beyhan, I was not there when this was discussed. Is this package needed or shall we provide the PR to remove it ?

beyhan:
Hi @a-hassanin,

IIRC @jpalermo wasn't sure whether it is needed for vSphere (see [this comment](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/328#issuecomment-1994192654)). I think the result of the discussion was to check on this. We can discuss this again during the FI WG todau.

beyhan:
We discussed during the FI WG meeting today and @jpalermo will try to find out whether it is still needed.

jpalermo:
I don't see it used anywhere in the agent or in the stemcell builder so I don't see how it could be used anywhere. I say lets remove it.

PR #419, opened 2025-03-11
Removes scsitools from the ubuntu noble.

ShilpaChandrashekara:
This commit removes scsitools from the ubuntu noble.

This is a part of cleaning up packages which are not really important during the lifecycle of a VM. As this is not used anymore in agent or in the stemcell builder.

Follow up on [415](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/415)

### DIFF
--- a/bosh-stemcell/spec/assets/dpkg-list-ubuntu.txt
+++ b/bosh-stemcell/spec/assets/dpkg-list-ubuntu.txt
@@ -436,7 +436,6 @@ rsyslog-gnutls
 rsyslog-openssl
 rsyslog-relp
 runit
-scsitools
 sed
 sensible-utils
 sg3-utils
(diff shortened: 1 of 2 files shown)


a-hassanin (changes requested):
I guess we need to remove it from https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/4ee8898fc6d5c8a68972e19228f6ba0295636f67/bosh-stemcell/spec/assets/dpkg-list-ubuntu.txt#L443 as well

ShilpaChandrashekara:
> I guess we need to remove it from
> 
> https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/4ee8898fc6d5c8a68972e19228f6ba0295636f67/bosh-stemcell/spec/assets/dpkg-list-ubuntu.txt#L443
> 
> as well

done

a-hassanin (approved):
LGTM

jpalermo (approved):

PR #425, opened 2025-05-08
fix: run logrotate once per minute

### LINKED ISSUE #424: Logrotate runs too infrequently
We have recently been running into issues where our disk filled up due to log files growing too fast before being compressed by logrotate. I noticed that the schedule for logrotate is coming directly from the stemcell:

https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/9321bcfb7b92c3c3b54ac118978ba00f12ebb493/stemcell_builder/stages/logrotate_config/assets/logrotate.timer#L1-L15

Is there a recommended way to increase the frequency? The frequency was originally once per minute but was changed to every 15min with c6b5705 to apparently align with the default of bionic. Nowadays the timer in ubuntu noble is a symlink to the [upstream example](https://github.com/logrotate/logrotate/blob/d94cc72dddbd2d195fddb158e07307415d2a22b1/examples/logrotate.timer) which sets it at once per day.

We would prefer something like every five minutes or even every minute, if there are no files to be rotated it shouldn't be too expensive to run logrotate frequently.

maxmoehl:
When a system is processing lots of data producing a high volume of logs running logrotate only once every 15 mintues can cause the disk to fill up if it is not big enough. This commit changes the logrotate timer to fire once per minute. If no logs need to be rotated logrotate will exit without modifying anything, this can be controlled via custom logrotate configuration.

This commit also removes the OnBootSec property as the timer fires every minute anyway, RandomizedDelaySec is reduced to one minute.

Resolves: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/424

### DIFF
--- a/stemcell_builder/stages/logrotate_config/assets/logrotate.timer
+++ b/stemcell_builder/stages/logrotate_config/assets/logrotate.timer
@@ -1,15 +1,10 @@
 [Unit]
-Description=random 1 min rotation of log files
+Description=Once per minute rotation of logfiles using logrotate.
 Documentation=man:logrotate(8) man:logrotate.conf(5)
 
 [Timer]
-OnBootSec=1min
-#This was once 'OnUnitActiveSec', but now is not. A Customer problem lead us to discover that it is said that
-#'oneshot' Units are never actually Active, so they will sometimes never be scheduled.
-#See: <https://github.com/systemd/systemd/issues/6680#issuecomment-326228938>
-OnCalendar=*-*-* *:00/15:00
-RandomizedDelaySec=2min
-Persistent=true
+OnCalendar=*-*-* *:*:00/5
+RandomizedDelaySec=1min
 
 [Install]
 WantedBy=timers.target


stemcell_builder/stages/logrotate_config/assets/logrotate.timer
```diff
@@ -1,15 +1,10 @@
...
 Documentation=man:logrotate(8) man:logrotate.conf(5)
 
 [Timer]
-OnBootSec=1min
-#This was once 'OnUnitActiveSec', but now is not. A Customer problem lead us to discover that it is said that
-#'oneshot' Units are never actually Active, so they will sometimes never be scheduled.
-#See: <https://github.com/systemd/systemd/issues/6680#issuecomment-326228938>
-OnCalendar=*-*-* *:00/15:00
-RandomizedDelaySec=2min
-Persistent=true
```
rkoster, line 12: Why did this line get removed?

maxmoehl: From systemd.timer(5):

> Takes a boolean argument. If true, the time when the service unit was last triggered is stored on disk. **When the timer is activated, the service unit is triggered immediately if it would have been triggered at least once during the time when the timer was inactive.** Such triggering is nonetheless subject to the delay imposed by RandomizedDelaySec=. This is useful to catch up on missed runs of the service when the system was powered down. Note that this setting only has an effect on timers configured with OnCalendar=. Defaults to false.

So this is only useful if you, for example, have daily triggers at 00:00 but your system might not be up at that time and you want the still fire the trigger the next time the system comes up. Since we run every minute that seems a bit pointless.

ramonskie:
i thinkthere was a reason why we had random time. as if it happens every minute with other tasks it could influence the host to much to interfere with other processes.

maxmoehl:
> i thinkthere was a reason why we had random time. as if it happens every minute with other tasks it could influence the host to much to interfere with other processes. 

I can also add back the randomness and set it to one minute if you prefer that.

xtreme-nitin-ravindran (dismissed):
lgtm

maxmoehl:
@rkoster @ramonskie could you please have another look?

ragaskar:
apologies, I have been slammed with other work and haven't found time to review. I've asked some colleagues to have a look.

aramprice:
This looks fine from a files-changed perspective, I am a little concerned about the frequency jump from `15 min` => `1 min`.

Is there evidence that this big of a jump is necessary? Would `10 min` or `5 min` be sufficient?

In the meeting we discussed there having been, perhaps in the Xenial or Trusty eras, issues with log rotate impacting performance. No one could remember the exact source of these issues though, unfortunately.

Has the `1 min` log rotation been run in your environment? If so did you experiment with any other log rotation frequencies?

maxmoehl:
> This looks fine from a files-changed perspective, I am a little concerned about the frequency jump from `15 min` => `1 min`.
> 
> Is there evidence that this big of a jump is necessary? Would `10 min` or `5 min` be sufficient?

This was just an initial proposal, I'm open to finding a middle-ground if you prefer a different value. 5 minutes should probably be often enough to prevent the full-disk issues we had in our envs.

> In the meeting we discussed there having been, perhaps in the Xenial or Trusty eras, issues with log rotate impacting performance. No one could remember the exact source of these issues though, unfortunately.
> 
> Has the `1 min` log rotation been run in your environment? If so did you experiment with any other log rotation frequencies?

As asked in [the issue](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/424) I'm not sure how I would adjust this easily without crafting a release which somehow overwrites the unit and reloads systemd. If there is a simple way that's good enough for me and we can keep the default at 15 minutes.

ramonskie:
i can live with 5 min on a random offset

maxmoehl:
> i can live with 5 min on a random offset

Perfect, I've adjusted the timer. I left `RandomizedDelaySec` at one minute and it should be noted that `AccuracySec` (currently) defaults to one minute. This allows the timer to be shifted up to one more minute to coalesce with some other activations. Example:

* OnCalendar would trigger at 10:05:00
* RandomizedDelay determines 10:05:43
* AccuracySec notices there's a timer scheduled at 10:06 and therefore coalesces the two, triggering logrotate at 10:06:00.

We could set `AccuracySec` to `1us` as suggest in the docs to prevent the last step from happening. See [systemd.timer(5)](https://man.archlinux.org/man/systemd.timer.5) for details.

maxmoehl:
@ramonskie @aramprice @xtreme-nitin-ravindran please have another look.

ramonskie (approved):
lgtm

aramprice (approved):

PR #428, opened 2025-05-14
Add yq package to docker image

rajathagasthya:

PR #430, opened 2025-05-14
Install yq from github on jammy

aramprice:
Resolves `Unable to locate package yq` error

This should resolve the error caused by https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/428


aramprice:
Closing in favor of https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/430

PR #434, opened 2025-06-17
fix: run logrotate once every 5 minutes

maxmoehl:
Same change as #425 but now for noble (minus a typo)

When a system is processing lots of data producing a high volume of logs running logrotate only once every 15 mintues can cause the disk to fill up if it is not big enough. This commit changes the logrotate timer to fire once per minute. If no logs need to be rotated logrotate will exit without modifying anything, this can be controlled via custom logrotate configuration.

This commit also removes the OnBootSec property as the timer fires every 5 minutes anyway, RandomizedDelaySec is reduced to one minute.

Resolves: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/424

### DIFF
--- a/stemcell_builder/stages/logrotate_config/assets/logrotate.timer
+++ b/stemcell_builder/stages/logrotate_config/assets/logrotate.timer
@@ -1,15 +1,10 @@
 [Unit]
-Description=random 1 min rotation of log files
+Description=Every 5 minutes rotation of logfiles using logrotate.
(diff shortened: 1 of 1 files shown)


ramonskie (approved):

ramonskie:
this can be merged as it already merged in jammy

benjaminguttmann-avtq (approved):

PR #439, opened 2025-08-29
Update Azure waagent version

HappyTobi:
Update the Waagent version that is used in the jammy stemcell.

The update is required because the current version is to old to get a update while runtime.


rkoster (approved):

a-hassanin (approved):
Looks fine. Thank you!

PR #443, opened 2025-10-10
Ignore SR-IOV VF interfaces for Azure stemcells

### LINKED ISSUE #366: MAC Address Detection Issues with Azure Accelerated Networking (SR-IOV Bonding)
**Problem Summary:**
The BOSH Agent's MAC address detector has a critical flaw when running on Azure VMs with Accelerated Networking enabled. The current implementation incorrectly handles [SR-IOV](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/overview-of-single-root-i-o-virtualization--sr-iov-) bonding scenarios, which can cause network connectivity issues during Azure live migrations.

**Background - Azure Accelerated Networking:**
When Accelerated Networking is enabled on Azure VMs, two network interfaces are created for each virtual NIC:
1. **Synthetic interface** (e.g., `eth0`) - uses the `hv_netvsc` driver
2. **SR-IOV Virtual Function (VF) interface** (e.g., `eth1` or `enP*`) - uses Mellanox `mlx4`/`mlx5` drivers

Both interfaces share the **same MAC address** and are automatically bonded by the `netvsc` driver. According to [Azure documentation](https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works#application-usage):

> **Applications should interact only with the synthetic interface**, just like in any other networking environment.

**Current Implementation Issue:**

The `linuxMacAddressDetector.DetectMacAddresses()` method in `mac_address_detector.go` scans all devices in `/sys/class/net/*` and builds a map where:
- **Key**: MAC address 
- **Value**: Interface name

When both synthetic and VF interfaces have the same MAC address, the map value gets overridden. Since the iteration order through `/sys/class/net/*` typically processes interfaces alphabetically, `eth1` (VF interface) often comes after `eth0` (synthetic interface), causing the map to incorrectly store the VF interface name as the final value.

**Impact:**
1. **Incorrect interface selection**: BOSH Agent may attempt to use the VF interface (`eth1`) instead of the synthetic interface (`eth0`)
2. **Live migration failures**: During Azure host servicing or live migrations, VF interfaces are temporarily removed, causing network disruption if applications depend on them.
3. **Reliability issues**: Applications interacting with VF interfaces directly violate Azure's networking model

**Expected Behavior:**
- For Azure environments, `eth0` (synthetic interface) should always be preferred over `eth1` (VF interface)

**Proposed Solutions:**
1. Option 1: Modify the mac_addres_detectory to filter out devices that have a `master` directory (typically the VF interface). This would ensure that our map of mac addresses only holds the synthetic interface.
2. Option 2: Add Azure-specific handling for accelerated networking scenarios.  Identify synthetic vs VF interfaces by checking for example the  `hv_netvsc` driver or interface flags.
3. Option 3: Avoid implementing IaaS-specific logic in the bosh-agent and instead configure an alias for the VF interface using `udev` rules within the bosh stemcell. The detector method already includes logic that reads from `ifalias`.

> [!NOTE]
> Here is an example how the interfaces can be distinguished based on the drivers:
> **Synthetic Interface**
>    ```sh
>    $ ethtool -i eth0 | grep driver
>    driver: hv_netvsc
>    ```
> **VF Interface**
>    ```sh
>    $ ethtool -i eth1 | grep driver
>    driver: mlx5_core
>    ```

**References:**
- [Azure Accelerated Networking Documentation](https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works)
- [Application Usage Guidelines](https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works#application-usage)

s4heid:
## Overview

When running Linux on cloud VMs with accelerated networking (SR-IOV), it’s critical to distinguish synthetic (paravirtualized) network interfaces from Virtual Function (VF) interfaces (direct hardware virtualization). This ensures the application uses the correct NIC (usually the synthetic one) for stable operation.

The goal of this change is to determine if an interface is a VF (hardware SR-IOV) or a synthetic interface and mark the VF interface with an alias.

## Accelerated Networking Differences on Cloud Providers

- **Azure**: On Azure both the synthetic NIC and the VF appear concurrently in the guest OS for the same network connection, sharing the same MAC address. The Linux `hv_netvsc` driver “bonds” the VF to the synthetic interface behind the scenes: it enslaves the VF under the synthetic device and automatically swaps the data path to the VF when available. The synthetic interface remains the primary one for configuration, while traffic is offloaded to the VF transparently. According to [Azure’s documentation][accelerated-networking-docs-azure],
  
    > “Applications should interact only with the synthetic interface… (the VF is under the control of the synthetic interface)”.

    If an application or agent mistakenly uses the VF directly, it can cause issues, especially since the VF may disconnect during host maintenance (e.g. Azure live migration) while the synthetic NIC stays up.

- **AWS**: On AWS, only one interface is visible to the guest for each ENI. If [enhanced networking (SR-IOV)][aws-enhanced-networking] is enabled, that interface itself uses the VF driver (e.g. `ixgbevf`, or `ena` on newer instances). If enhanced networking is disabled or not supported, the same single interface uses a synthetic driver (Xen vif in the example). There is no second “paired” interface with the same MAC on AWS — the NIC is either synthetic or VF, not both at once.

## Solution

1. Use **udev** to mark the VF as unmanaged (this handles the OS/network config side – ensuring only the synthetic NIC is configured and considered "online").
2. Configure **systemd-networkd** to skip devices marked with `AZURE_UNMANAGED_SRIOV=1` by udev.
3. Use **udev** to set an `ifalias` on the VF as a clear indicator (especially for the  bosh-agent, and for human debugging).

> [!NOTE]
> Marking unmanaged ensures that no automatic scripts or services will configure the VF. It will not get an IP via DHCP, and it should not be waited on during boot (preventing delays). For instance, a [Red Hat bug report][redhat-bug] noted that without the rule, `NetworkManager-wait-online.service` could hang for ~90 seconds on Azure while waiting for the VF interface to be up.
>
> Marking an interface unmanaged by NM/networkd does not remove it from the kernel. It still exists and shows up in `/sys/class/net`. So a user-space program (like BOSH agent) that directly scans system interfaces will still see it. BOSH agent does not ask NetworkManager which interfaces are managed; it just looks at the NICs present and their MACs. Thus, unmanaged marking alone doesn’t stop BOSH from confusing the VF for the primary, but combined with the agent’s [updated logic][bosh-agent-change] it will use the correct NIC.

## How to validate the solution works?

**Prerequisites**:

1. BOSH Director on Azure
2. Custom ubuntu-jammy stemcell which includes the [`fix`][stemcell-change] in bosh-linux-stemcell-builder and the [`fix`][bosh-agent-change] in the bosh-agent
3. Custom stemcell uploaded to the director
4. Test deployment referencing the custom stemcell and accelerated networking enabled

**What you should observe**:

- On the VMs you should see that in the `/sys/class/net/` directory, both interfaces are present, `eth0` and `eth1`. 

    ```sh
    $ ls /sys/class/net/ | grep eth
      eth0
      eth1
    ```

- Using `ethtool -i <interface name> | grep driver` command you can find out which one is the synthetic interface, based on the drivers. If the driver is hv_netvsc, it's the synthetic interface. The VF interface has a driver name that contains "mlx." The VF interface is also identifiable because its flags field includes `SLAVE`.

    ```sh
    $ ethtool -i eth0 | grep driver
      driver: hv_netvsc

    $ ethtool -i eth1 | grep driver
      driver: mlx5_core

    $ ip link | grep SLAVE
      3: eth1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master eth0 state UP mode DEFAULT group default qlen 1000

- run `ip link` to verify the VF has the alias set:

    ```sh
    $ ip link | grep alias
      alias sriov-vf
    ```

- When you identified the synthetic interface, let's say it's `eth0`, make sure there is a systemd-networkd config in `/etc/systemd/network/` for that interface.

    ```sh
    $ ls /etc/systemd/network/ | grep eth0
      10_eth0.network
    ```

- You can also verify in the bosh-agent logs that the bosh-agent ignores the VF interface:

    ```sh
    $ zgrep "sriov-vf" /var/vcap/bosh/log/*
      /var/vcap/bosh/log/@4000000068e933d41e319b4c.s:2025-10-10_16:26:48.49426 sriov-vf
      /var/vcap/bosh/log/@4000000068e933d41e319b4c.s:2025-10-10_16:26:48.49427 [MacAddressDetector] 2025/10/10 16:26:48 DEBUG - Ignoring SR-IOV VF interface: eth1 (ifalias: sriov-vf)
    ```

## References

- [azure-vm-utils][az-vm-utils] - this package uses udev rules to mark Azure VFs similarly to the proposal

## Related

- Resolves cloudfoundry/bosh-agent#366
- Relates to cloudfoundry/bosh-agent#374

[stemcell-change]: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/532b8f6cda1ea59c7f71989a505207771ab87c2f
[bosh-agent-change]: https://github.com/cloudfoundry/bosh-agent/pull/374
[redhat-bug]: https://issues.redhat.com/browse/RHEL-15222
[aws-enhanced-networking]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sriov-networking.html
[az-vm-utils]: https://github.com/Azure/azure-vm-utils/tree/main
[accelerated-networking-docs-azure]: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works#application-usage

### DIFF
--- a/bosh-stemcell/spec/stemcells/azure_spec.rb
+++ b/bosh-stemcell/spec/stemcells/azure_spec.rb
@@ -42,4 +42,41 @@ describe 'Azure Stemcell', stemcell_image: true do
       its(:content) { should include('"PartitionerType": "parted"') }
     end
   end
+
+  context 'installed by system_azure_network', {
+    exclude_on_alicloud: true,
+    exclude_on_aws: true,
+    exclude_on_google: true,
+    exclude_on_vcloud: true,
+    exclude_on_vsphere: true,
+    exclude_on_warden: true,
+    exclude_on_openstack: true,
+    exclude_on_softlayer: true,
+  } do
+    describe 'SR-IOV VF udev rules' do
+      subject { file('/etc/udev/rules.d/10-azure-sriov-unmanaged.rules') }
+
+      it { should be_mode(644) }
+      it { should be_owned_by('root') }
+
+      its(:content) { should match /SUBSYSTEM=="net"/ }
+      its(:content) { should match /ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/ }
+      its(:content) { should match /ENV\{AZURE_UNMANAGED_SRIOV\}="1"/ }
+      its(:content) { should match /ENV\{ID_NET_MANAGED_BY\}="unmanaged"/ }
+      its(:content) { should match /ENV\{NM_UNMANAGED\}="1"/ }
+      its(:content) { should match /ATTR\{ifalias\}="sriov-vf"/ }
+    end
+
+    describe 'systemd network configuration for unmanaged SR-IOV devices' do
+      subject { file('/etc/systemd/network/01-azure-sriov-unmanaged.network') }
+
+      it { should be_mode(644) }
+      it { should be_owned_by('root') }
+
+      its(:content) { should match /\[Match\]/ }
+      its(:content) { should match /Property=AZURE_UNMANAGED_SRIOV=1/ }
+      its(:content) { should match /\[Link\]/ }
+      its(:content) { should match /Unmanaged=yes/ }
+    end
+  end
 end
--- a/stemcell_builder/stages/system_azure_network/apply.sh
+++ b/stemcell_builder/stages/system_azure_network/apply.sh
@@ -20,3 +20,11 @@ EOS
 # The port 65330 is unusable on Azure
 cp $dir/assets/90-azure-sysctl.conf $chroot/etc/sysctl.d
 chmod 0644 $chroot/etc/sysctl.d/90-azure-sysctl.conf
+
+# Install SR-IOV VF udev rule to mark network interfaces as unmanaged and set ifalias
+cp $dir/assets/10-azure-sriov-unmanaged.rules $chroot/etc/udev/rules.d
+chmod 0644 $chroot/etc/udev/rules.d/10-azure-sriov-unmanaged.rules
+
+# Apply the systemd.network configuration, to ignore unmanaged devices
+cp $dir/assets/01-azure-sriov-unmanaged.network $chroot/etc/systemd/network
+chmod 0644 $chroot/etc/systemd/network/01-azure-sriov-unmanaged.network
--- a/stemcell_builder/stages/system_azure_network/assets/01-azure-sriov-unmanaged.network
+++ b/stemcell_builder/stages/system_azure_network/assets/01-azure-sriov-unmanaged.network
@@ -1,0 +1,11 @@
+# Azure VMs with accelerated networking may have MANA, mlx4, or mlx5 SR-IOV
+# devices which are transparently bonded to a synthetic hv_netvsc device.
+#
+# 10-azure-unmanaged-sriov.rules will mark these devices with
+# AZURE_UNMANAGED_SRIOV=1, allowing this configuration to set them as unmanaged.
+
+[Match]
+Property=AZURE_UNMANAGED_SRIOV=1
+
+[Link]
+Unmanaged=yes
--- a/stemcell_builder/stages/system_azure_network/assets/10-azure-sriov-unmanaged.rules
+++ b/stemcell_builder/stages/system_azure_network/assets/10-azure-sriov-unmanaged.rules
@@ -1,0 +1,19 @@
+# Azure VMs with accelerated networking may have MANA, mlx4, or mlx5 SR-IOV devices
+# which are transparently bonded to a synthetic hv_netvsc device.
+#
+# Mark devices with the IFF_SLAVE bit set as unmanaged devices:
+#   AZURE_UNMANAGED_SRIOV=1:     Custom tag to mark the interface as unmanaged in systemd-networkd (01-azure-sriov-unmanaged.network).
+#   ID_NET_MANAGED_BY=unmanaged: Standard tag for systemd-networkd (v255+).
+#   NM_UNMANAGED=1:              Tells NetworkManager to ignore the interface.
+#
+# Additionally, set an ifalias to allow the bosh-agent to ignore the device, as both
+# interfaces, the synthetic and the VF interface, share the same MAC address and only
+# the synthetic interface should be used by the application.
+
+SUBSYSTEM=="net", ACTION!="remove", KERNEL!="lo", \
+    DRIVERS=="mana|mlx4_core|mlx5_core", \
+    ATTR{flags}=="0x?[89ABCDEF]*", \
+    ENV{AZURE_UNMANAGED_SRIOV}="1", \
+    ENV{ID_NET_MANAGED_BY}="unmanaged", \
+    ENV{NM_UNMANAGED}="1", \
+    ATTR{ifalias}="sriov-vf"


fmoehler (approved):

aramprice (approved):

aramprice:
Candidate for https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/450

selzoc:
FYI @s4heid this change is causing the azure stemcell build to fail: https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/stemcells-ubuntu-jammy/jobs/build-azure-hyperv/builds/430

```
Failures:

  1) Azure Stemcell installed by system_azure_network SR-IOV VF udev rules is expected to be mode 644
     Failure/Error: it { should be_mode(644) }
       expected `#<ShelloutTypes::File:0x00007fcbeb892a38 @path="/etc/udev/rules.d/10-azure-sriov-unmanaged.rules", @chroot=#<ShelloutTypes::Chroot:0x00007fcbeb892a60>>.mode?(644)` to be truthy, got false
     # ./spec/stemcells/azure_spec.rb:59:in `block (4 levels) in <top (required)>'

  2) Azure Stemcell installed by system_azure_network SR-IOV VF udev rules content is expected to match /ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/
     Failure/Error: its(:content) { should match /ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/ }
     
       expected "# Azure VMs with accelerated networking may have MANA, mlx4, or mlx5 SR-IOV devices\n# which are transparently bonded to a synthetic hv_netvsc device.\n#\n# Mark devices with the IFF_SLAVE bit set as unmanaged devices:\n#   AZURE_UNMANAGED_SRIOV=1:     Custom tag to mark the interface as unmanaged in systemd-networkd (01-azure-sriov-unmanaged.network).\n#   ID_NET_MANAGED_BY=unmanaged: Standard tag for systemd-networkd (v255+).\n#   NM_UNMANAGED=1:              Tells NetworkManager to ignore the interface.\n#\n# Additionally, set an ifalias to allow the bosh-agent to ignore the device, as both\n# interfaces, the synthetic and the VF interface, share the same MAC address and only\n# the synthetic interface should be used by the application.\n\nSUBSYSTEM==\"net\", ACTION!=\"remove\", KERNEL!=\"lo\", \\\n    DRIVERS==\"mana|mlx4_core|mlx5_core\", \\\n    ATTR{flags}==\"0x?[89ABCDEF]*\", \\\n    ENV{AZURE_UNMANAGED_SRIOV}=\"1\", \\\n    ENV{ID_NET_MANAGED_BY}=\"unmanaged\", \\\n    ENV{NM_UNMANAGED}=\"1\", \\\n    ATTR{ifalias}=\"sriov-vf\"\n" to match /ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/
       Diff:
       @@ -1,19 +1,37 @@
       -/ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/
       +# Azure VMs with accelerated networking may have MANA, mlx4, or mlx5 SR-IOV devices
       +# which are transparently bonded to a synthetic hv_netvsc device.
       +#
       +# Mark devices with the IFF_SLAVE bit set as unmanaged devices:
       +#   AZURE_UNMANAGED_SRIOV=1:     Custom tag to mark the interface as unmanaged in systemd-networkd (01-azure-sriov-unmanaged.network).
       +#   ID_NET_MANAGED_BY=unmanaged: Standard tag for systemd-networkd (v255+).
       +#   NM_UNMANAGED=1:              Tells NetworkManager to ignore the interface.
       +#
       +# Additionally, set an ifalias to allow the bosh-agent to ignore the device, as both
       +# interfaces, the synthetic and the VF interface, share the same MAC address and only
       +# the synthetic interface should be used by the application.
       +
       +SUBSYSTEM=="net", ACTION!="remove", KERNEL!="lo", \
       +    DRIVERS=="mana|mlx4_core|mlx5_core", \
       +    ATTR{flags}=="0x?[89ABCDEF]*", \
       +    ENV{AZURE_UNMANAGED_SRIOV}="1", \
       +    ENV{ID_NET_MANAGED_BY}="unmanaged", \
       +    ENV{NM_UNMANAGED}="1", \
       +    ATTR{ifalias}="sriov-vf"
       
     # ./spec/stemcells/azure_spec.rb:63:in `block (4 levels) in <top (required)>'

  3) Azure Stemcell installed by system_azure_network systemd network configuration for unmanaged SR-IOV devices is expected to be mode 644
     Failure/Error: it { should be_mode(644) }
       expected `#<ShelloutTypes::File:0x00007fcbeb7de4c0 @path="/etc/systemd/network/01-azure-sriov-unmanaged.network", @chroot=#<ShelloutTypes::Chroot:0x00007fcbeb7de4e8>>.mode?(644)` to be truthy, got false
     # ./spec/stemcells/azure_spec.rb:73:in `block (4 levels) in <top (required)>'

Finished in 49.16 seconds (files took 0.98979 seconds to load)
544 examples, 3 failures, 1 pending

Failed examples:

rspec ./spec/stemcells/azure_spec.rb:59 # Azure Stemcell installed by system_azure_network SR-IOV VF udev rules is expected to be mode 644
rspec ./spec/stemcells/azure_spec.rb:63 # Azure Stemcell installed by system_azure_network SR-IOV VF udev rules content is expected to match /ATTR\{flags\}=="0\?\?\[89ABCDEF\]\*"/
rspec ./spec/stemcells/azure_spec.rb:73 # Azure Stemcell installed by system_azure_network systemd network configuration for unmanaged SR-IOV devices is expected to be mode 644
```

PR #442, opened 2025-10-07
Improve Azure chrony accuracy: change Hyper-V PTP poll 3 -> -1, add RT scheduling + OOM hardening

naioja:
Switch chrony PHC refclock polling for /dev/ptp_hyperv from poll 3 to poll -1 (with existing dpoll -2) to tighten synchronization on Azure VMs. Added a systemd drop-in granting FIFO scheduling (priority 50) and OOMScoreAdjust -500 so chronyd remains stable under memory pressure.

**Background:** 
poll 3 corresponds to a 2^3 = 8s base interval; with dpoll -2 the dynamic range was 2–8s. This slower sampling delayed convergence after VM start, suspend/resume, or host migration and increased offset/jitter under bursty load. Setting poll -1 (2^-1 = 0.5s) shifts dynamic range (with dpoll -2) to 0.125–0.5s, yielding faster lock and lower RMS offset for time-sensitive workloads (e.g. distributed tracing, log ordering, certificate validation windows).

**Changes:**
Refclock line: poll 3 -> poll -1 (kept dpoll -2, offset 0, stratum 2)
Added systemd chrony.service override: FIFO scheduling, priority 50, OOMScoreAdjust -500.

**Benefits:**
Quicker convergence after boot or host events.
Reduced steady-state offset and jitter on Azure Hyper-V PTP.
Increased resilience against OOM kills and CPU contention.

**Testing Performed:**

- Long-running observation: stable lower RMS offset and reduced jitter versus previous config
- chronyc tracking showed improved convergence after restart events or live migrations
- Multi-day fio mixed read/write stress (sustained high IO + CPU bursts) showed no degradation in chrony accuracy, no adverse impact on workload performance, and no elevated chronyd CPU usage

- Verified chronyc tracking: reduced offset (avg improvement ~4 ms -> ~50 µs) and lower jitter.
image

- VS. default chrony configuration
image

- Monitored CPU: negligible (<0.1% increase) on tested Standard D2as v5 (2 vcpus, 8 GiB memory).
image

- VS. default chrony configuration
image

Confirmed systemd status and real-time scheduling applied :

```bash
systemctl show chrony | egrep '(CPUSchedulingPriority|CPUSchedulingPolicy|OOMScoreAdjust)'
OOMScoreAdjust=-500
CPUSchedulingPolicy=1
CPUSchedulingPriority=50
```

**Rollback:** 
Change **poll -1 back to 3 and remove the systemd drop-in** directory if higher sampling cost becomes problematic.

[...]

PR #445, opened 2025-10-17
Fix path to chrony.service

s4heid:
This change sets the correct path when creating the chrony.service directory. The wrong path was introduced in #442.


ramonskie (approved):
fixes ci :)

PR #449, opened 2025-10-23
Disable unnecessary floppy kernel module

s4heid:
This PR removes loading of the legacy floppy module from the stemcell build.

Rationale:

- Module is obsolete hardware support; keeping it expands kernel attack surface (extra code paths, potential low‑severity but chronic CVEs).
- Azure (including Hyper-V) presents no virtual floppy device; upstream guidance (Microsoft LISA [core test suite](https://github.com/microsoft/lisa/blob/main/lisa/microsoft/testsuites/core/floppy.py)) explicitly expects/assumes floppy is absent.
- Eliminates pointless udev/probing delays and reduces initramfs/kernel footprint for slightly faster, cleaner boot.

### DIFF
--- a/bosh-stemcell/spec/support/os_image_linux_kernel_modules_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_linux_kernel_modules_shared_examples.rb
@@ -83,4 +83,10 @@ shared_examples_for 'a Linux kernel module configured OS image' do
       its(:content) { should match 'install rds /bin/true' }
     end
   end
+
+  context 'prevent floppy module from being loaded' do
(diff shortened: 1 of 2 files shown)


neddp (approved):

Alphasite:
I don’t know if this is still true but the vsphere cpi used this module for its equivalent of cloud init. It may have moved to virtual CDs but we need to double check this.

selzoc:
Pretty sure the vsphere cpi uses a virtual cd drive, at least according to https://bosh.io/docs/vsphere/#concepts

rkoster (approved):

PR #452, opened 2025-11-06
Update stemcell-repack-debugger

ionphractal:
While addressing https://github.com/cloudfoundry/bosh-agent/issues/337 I was in the need to build stemcells for Jammy and was limited in where I build them (with limited disk space). Hence I had to add/adjust a few things to this script.

* Add support for both jammy and noble
* Add version bumping to a specific version
* Add env var for specifying a temp path in case your normal temp directory is too small
* Add env var for specifying an output path for the stemcell
* Add automatic version bumping if the output stemcell exists
* Cleanup more directories when exitting via trap
* Speed up version detection by specifying the order of files to be compressed


ramonskie:
i worked on this a while ago but never made a PR see https://github.com/cloudfoundry/bosh-linux-stemcell-builder/tree/helper-script

Please note that warden stemcells are packaged differently

aramprice (dismissed):

scripts/stemcell-repack-debugger.sh
```diff
@@ -48,8 +52,13 @@ for arg in "$@"; do
             convert_cert_file="$2"
             shift 2
             ;;
-        --bump-version)
+        --bump-version*)
             bump_version=true
+            if [ -n "$OUT" -a -f "$OUT" ]; then
+                version=$(tar xf "$OUT" stemcell.MF --occurrence=1 -O | yq .version)
```
Alphasite: Can you use bosh interpolate instead of yq here? In general I’d like to avoid adding additional dependencies

ionphractal:
@Alphasite replaced all yq usage now.
@ramonskie would've been nice to have your version earlier. Maybe we should use the time to integrate all in one improvement... though it is hard for me to spare that atm and I'm on vac the next 2 weeks.

rkoster (approved):

beyhan (approved):

beyhan:
We discussed this during the FI WG meeting today and decided that the improvements could be implemented in a second pr following this.

PR #454, opened 2025-11-10
Prevent kernel from auto-loading floppy

s4heid:
In a previous change (#449), loading the floppy module was disabled by redirecting the floppy command to true. However, Buffer I/O errors observed in the `dmesg` output indicate that the kernel is still attempting to load the floppy module when the hardware supports it.

```txt
blk_update_request: I/O error, dev fd0, sector 0 op 0x0:(READ) flags 0x80700 phys_seg 1 prio class 0
floppy0: disk absent or changed during operation
```

This suggests that during boot:
1. Kernel detects floppy controller hardware
2. udev/kernel auto-loads floppy module (install directive not active yet)
3. Floppy driver starts, finds no disk → I/O errors
4. install directive becomes active (too late!)

The blacklist in /etc/modprobe.d only affects modprobe after the root filesystem is active. If the initramfs contains a `floppy.ko` file and is not rebuilt after the blacklist has been applied, the initramfs auto-loads the floppy driver.
`update-initramfs` rebuilds the initramfs and includes the /etc/modprobe.d rules to the root fs.

PR #455, opened 2025-11-14
Mount /sys in system_kernel_modules for initramfs generation

s4heid:
The system_kernel_modules stage fails when running update-initramfs with FIPS kernels because the FIPS initramfs hooks require `/sys` to be mounted for hardware introspection when MODULES=dep is configured.

The `run_in_chroot` helper creates an isolated mount namespace via `unshare -m`, which does not mount `/sys`. This causes mkinitramfs to fail with "mkinitramfs: MODULES dep requires mounted sysfs on /sys" when the FIPS hooks attempt to scan hardware capabilities.

Mount `/sys` explicitly in `system_kernel_modules/apply.sh` before calling update-initramfs to ensure the kernel device model and driver information is available for initramfs generation.

#### Related

- https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/454#issuecomment-3530504160

#### Alternative Solutions

We could also mount `/sys` directly in the `run_in_chroot` helper, as `/sys` is also mounted in other places:

```
❯ grep -r "mount.*bind.*sys" stemcell_builder/stages/
stemcell_builder/stages/base_fips_apt/apply.sh:mount --bind /sys "$chroot/sys"
stemcell_builder/stages/system_fips_kernel/apply.sh:mount --bind /sys "$chroot/sys"
stemcell_builder/stages/restore_apt_sources/apply.sh:mount --bind /sys "$chroot/sys"
stemcell_builder/stages/base_apt/apply.sh:mount --bind /sys "$chroot/sys"
```

### DIFF
--- a/stemcell_builder/stages/system_kernel_modules/apply.sh
+++ b/stemcell_builder/stages/system_kernel_modules/apply.sh
@@ -31,4 +31,6 @@ alias lbm-nouveau off' >> $chroot/etc/modprobe.d/blacklist-nouveau.conf
 
 rm -rf $chroot/lib/modules/*/kernel/zfs $chroot/usr/src/linux-headers-*/zfs
 
+mount --bind /sys "$chroot/sys"
+add_on_exit "umount $chroot/sys"
 run_in_chroot $chroot "update-initramfs -u -k all"


aramprice (approved):

aramprice:
@s4heid - thank you for the quick turn around, this fix makes sense to me.

PR #460, opened 2025-12-01
Fix race condition causing sshd start failure during provisioning

### LINKED ISSUE #458: ssh.service failures on azure stemcells
Occassionally, it happens that cloud-init reports issues regarding the startup of `ssh.service`. In the cloud-init log you would see the following error:

```sh
$ cloud-init status --long
status: done
extended_status: degraded done
boot_status_code: enabled-by-generator
last_update: Thu, 01 Jan 1970 00:00:27 +0000
detail: DataSourceAzure [seed=/dev/sr0]
errors: []
recoverable_errors:
WARNING:
	- Unhandled non-multipart (text/x-not-multipart) userdata: 'b'{"server":{"name":"agent'...'
	- 'ssh_pwauth' configuration may not be applied. Cloud-init was unable to restart SSH daemon due to error: 'Unexpected error while running command. Command: ['systemctl', 'restart', 'ssh', '--job-mode=ignore-dependencies'] Exit code: 1 Reason: - Stdout:  Stderr: Job for ssh.service failed because the control process exited with error code.         See "systemctl status ssh.service" and "journalctl -xeu ssh.service" for details.'
```

The journal log shows that the host keys could not be loaded, causing `ssh.service` to fail with exit status 1.

```sh
systemd[1]: Starting OpenBSD Secure Shell server...
░░ Subject: A start job for unit ssh.service has begun execution
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ A start job for unit ssh.service has begun execution.
░░ 
░░ The job identifier is 307.
sshd[928]: Unable to load host key: /etc/ssh/ssh_host_rsa_key
sshd[928]: Unable to load host key: /etc/ssh/ssh_host_ecdsa_key
sshd[928]: Unable to load host key: /etc/ssh/ssh_host_ed25519_key
sshd[928]: sshd: no hostkeys available -- exiting.
systemd[1]: ssh.service: Control process exited, code=exited, status=1/FAILURE
░░ Subject: Unit process exited
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ An ExecStartPre= process belonging to unit ssh.service has exited.
░░ 
░░ The process' exit code is 'exited' and its exit status is 1.
Nov 06 10:12:01 fb126bb7-ad45-4af4-a70e-8da723c21826 systemd[1]: ssh.service: Failed with result 'exit-code'.
░░ Subject: Unit failed
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ The unit ssh.service has entered the 'failed' state with result 'exit-code'.
systemd[1]: Failed to start OpenBSD Secure Shell server.
░░ Subject: A start job for unit ssh.service has failed
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ A start job for unit ssh.service has finished with a failure.
```

Looking at the serial console of the VM, we are pointed to a file in cloud-init, [`cc_set_passwords.py`](https://github.com/canonical/cloud-init/blob/987b770af415f2bb7e26fec0abad746b3f0186fc/cloudinit/config/cc_set_passwords.py#L119-L120):

```sh
[  OK  ] Started Run anacron jobs.
[   12.661141] cloud-init[912]: Cloud-init v. 25.2-0ubuntu1~22.04.1 running 'modules:config' at Mon, 17 Nov 2025 16:54:26 +0000. Up 12.58 seconds.
[  OK  ] Stopped OpenBSD Secure Shell server.
         Starting OpenBSD Secure Shell server...
[  OK  ] Stopped OpenBSD Secure Shell server.
         Starting OpenBSD Secure Shell server...
[FAILED] Failed to start OpenBSD Secure Shell server.
See 'systemctl status ssh.service' for details.
[  OK  ] Started Dispatcher daemon for systemd-networkd.
[   12.721602] cloud-init[912]: 2025-11-17 16:54:26,851 - cc_set_passwords.py[WARNING]: 'ssh_pwauth' configuration may not be applied. Cloud-init was unable to restart SSH daemon due to error: 'Unexpected error while running command.
[   12.729634] cloud-init[912]: Command: ['systemctl', 'restart', 'ssh', '--job-mode=ignore-dependencies']
[   12.733487] cloud-init[912]: Exit code: 1
[   12.735251] cloud-init[912]: Reason: -
[   12.736812] cloud-init[912]: Stdout:
[   12.738550] cloud-init[912]: Stderr: Job for ssh.service failed because the control process exited with error code.
[   12.743876] cloud-init[912]:         See "systemctl status ssh.service" and "journalctl -xeu ssh.service" for details.'
[  OK  ] Finished Cloud-init: Config Stage.
[  OK  ] Stopped OpenBSD Secure Shell server.
         Starting OpenBSD Secure Shell server...
[FAILED] Failed to start OpenBSD Secure Shell server.
See 'systemctl status ssh.service' for details.
[  OK  ] Stopped OpenBSD Secure Shell server.
         Starting OpenBSD Secure Shell server...
[FAILED] Failed to start OpenBSD Secure Shell server.
See 'systemctl status ssh.service' for details.
[  OK  ] Stopped OpenBSD Secure Shell server.
[FAILED] Failed to start OpenBSD Secure Shell server.
```

It looks like this is a recoverable error, as the ssh.service is active and running on the VM.

```sh
$ systemctl status ssh.service
● ssh.service - OpenBSD Secure Shell server
     Loaded: loaded (/lib/systemd/system/ssh.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2025-11-17 16:50:28 UTC; 1 week 0 days ago
       Docs: man:sshd(8)
             man:sshd_config(5)
   Main PID: 4066 (sshd)
      Tasks: 1 (limit: 9431)
     Memory: 3.9M
     CGroup: /system.slice/ssh.service
             └─4066 "sshd: /usr/sbin/sshd -D [listener] 0 of 10-100 startups"
```

However, it appears that, besides cloud-init, another process might also be attempting to manage the SSH service at the same time, which occasionally results in these errors.

s4heid:
* Run first-boot tasks via systemd so sshd never races with host-key regeneration. The old `rc.local` script ran after network.target, but in parallel with other regular system services, like ssh.service. Therefore, ssh.service often started (and restarted) while `/root/firstboot.sh` was deleting keys. cloud-init’s set-passwords module made this worse by restarting ssh mid-run.
  * Replace `rc.local` with a oneshot firstboot.service (delete keys, create new keys, reconfigure sysstat) that runs Before=ssh.service and leaves the `/root/firstboot_done` file as a marker.
  * Add a cloud-config.service drop-in so cloud-init's config stage waits for firstboot.service, and
  * Update walinuxagent.service to wait for firstboot.service, ensuring ssh keys have been regenerated. This guarantees sshd, cloud-init, and WALinuxAgent all start only after the first-boot tasks succeed.

Resolves #458


s4heid:
> [!WARNING]
> It's important to be aware that this change could affect how the ssh service behaves. If the firstboot script was intended **only** for host key regeneration, using the `ssh-keygen -A` command should be sufficient.
> However, if we want to continue using the `dpk-reconfigure` command, it becomes a bit more challenging to ensure that firstboot.service runs before ssh.service, since `dpk-reconfigure` attempts to restart ssh.service, which would result in a deadlock. One possible solution would be to do something like this:

```
[Service]
Type=oneshot
ExecStartPre=/bin/sh -c '{ echo \"#!/bin/sh\nexit 101\"; } > /usr/sbin/policy-rc.d && chmod +x /usr/sbin/policy-rc.d'
ExecStartPre=/bin/sh -c '/bin/rm -f /etc/ssh/ssh_host*key*'
ExecStart=/usr/sbin/dpkg-reconfigure -fnoninteractive -pcritical openssh-server
ExecStart=/usr/sbin/dpkg-reconfigure -fnoninteractive sysstat
ExecStartPost=/bin/rm -f /usr/sbin/policy-rc.d
```

ramonskie:
we should not introduce this within jammy.
as it will break warden stemcells. because if the host does not have systemd the vm with the jammy stemcell will fail.

we currently have similar  issues on noble as well as we have set bosh-agent to use systemd

s4heid:
Thank you for the context, @ramonskie. I didn't realize that warden stemcells wouldn't be compatible with this. Is noble affected by the same issue? If we could get this fixed at least for noble, I'd be happy.. :)

aramprice:
@s4heid - we changed the branch to `ubuntu-noble` but perhaps a better approach would be to cherry-pick these changes onto an `ubuntu-noble` based branch.

s4heid:
thanks @aramprice! I rebased onto noble and validated that the change still works.

ystros (approved):
Looks reasonable to me, barring any additional concerns around warden.

aramprice:
This may be causing a regression
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/485

PR #461, opened 2025-12-04
Update waagent to v2.15.0.1

s4heid:
* upgrades waagent to the latest version
 * syncs the systemd config, which adds memory-accounting and a dependency
   to cloud-init.service, which ensures that the waagent.service does
   not start before cloud-init has finished.

<https://github.com/Azure/WALinuxAgent/blob/v2.15.0.1/init/ubuntu/walinuxagent.service>


ystros (approved):

aramprice:
Thanks @s4heid !

PR #464, opened 2025-12-22
Add vsphere guestinfo settings source to jammy stemcell

julian-hj:
This PR adds a VsphereGuestInfo settings source, causing the agent to check the new settings source prior to trying the CDROM settings source.

It should be safe to merge now that the [corresponding bosh agent](https://github.com/cloudfoundry/bosh-agent/pull/392) PR is merged, and has been promoted in CI.


Alphasite (approved):
Assuming the agent changes ar emerged, this makes sense.

PR #467, opened 2026-01-27
Add VsphereGuestInfo settings source to noble stemcell

julian-hj:
This allows the agent to fetch settings from guestinfo in vmware rpc in vsphere instead of mounting the CD-ROM. There is already a promoted bosh-agent available, so this change should be ready to merge.


Alphasite (approved):
Assuming the agent changes are merged, this makes sense.

aramprice (approved):

ystros (approved):

PR #470, opened 2026-02-09
Remove monit permit access and NATS firewall setup

rkoster:
## Summary

- Remove cgroup v1 net_cls-based monit API access control mechanism
- Remove monit wrapper script, helper functions, and iptables rules
- Remove permit_monit_access call from agent startup

The monit binary now runs directly without a wrapper. Access control will be managed by the bosh-agent's internal firewall implementation.

This prepares the stemcell builder for the bosh-agent changes that move firewall management into the agent itself.

## Related

- cloudfoundry/bosh-agent#399
- cloudfoundry/pxc-release#97 (pxc-release was the only consumer of permit_monit_access)


mariash (approved):

aramprice (approved):

rkoster:
Don't merge before: https://github.com/cloudfoundry/bosh-agent/pull/399

abg:
Per the comment from @colins from [the related pxc-release PR](https://github.com/cloudfoundry/pxc-release/pull/97#issuecomment-3899402077)

> @rkoster there is some more work to do for this. Specifically for backward compatibility for some MySQL components.  We'd need some like monit-access-helper.sh in jammy so that we don't break features for customers.

Adding some color here:

This `monit-access-helper.sh` pattern is apparently being used by some of our internal bosh releases - this usually involves some niche database workflows that do some in-place restore like stopping the database service, putting some files in places and starting the service again. 

Would it be acceptable if we kept `monit-access-helper.sh` for the ubunt-jammy stemcell, but reimplemented in terms of the nftables work?   Maybe pulling in some of the work from https://github.com/cloudfoundry/pxc-release/pull/97 into bosh-agent and adding some additional `/var/vcap/bosh/bin/bosh-monit-access` helper?

That may make sense as a follow up backwards compatibility PR.

rkoster:
Yes I'm not against backward compatibility, it is just out of scope for the use case that I'm interested in so me personally won't be doing the work. That being said I'm happy to review a PR adding backward compatibility.

aramprice:
I think for backwards compatibility that `stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh` can not be removed because older pxc-releases will not function without this file.

I propose we maove `bosh-monit-access` from https://github.com/cloudfoundry/pxc-release/pull/97 into the stemcell and hollow out `stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh` so that it calls this new binary (on newer stemcells). This leaves the existing script in place which allows older releases to be deployed on newer Jammy stemcells.

cc @mariash

rkoster:
Could work, but needs a bit more work. Because the pxc helper script is not generic in its current form. It is also difficult to figure out the right cgroup etc. So it's probably better to just allow access for vcap user when the script would be called on jammy.

PR #473, opened 2026-02-12
Prevent systemd-binfmt from running in containers

KauzClay:
extracting from https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/468#issuecomment-3891802803

When stemcells run as privileged containers (e.g., Docker CPI on Apple Silicon), systemd-binfmt clears the host's binfmt_misc registrations, including Rosetta, causing "exec format error" for x86_64 processes.

Add a drop-in override with ConditionVirtualization=!container to skip the service in containers while preserving normal behavior on VMs.

related to https://github.com/cloudfoundry/bosh-deployment/issues/497


mkocher:
I've merged #500, which replaced this with the same change made on the Noble branch

PR #474, opened 2026-02-18
Use bosh-agent enable-monit-access if available

mariash:
monit-access-helper will use "enable-monit-access" bosh-agent command if it is available and nftables tool is present. Otherwise fall back to old monit cgroup access setup.

This PR requires https://github.com/cloudfoundry/bosh-agent/pull/406

### DIFF
--- a/stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh
+++ b/stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh
@@ -1,0 +1,3 @@
+permit_monit_access() {
+  /var/vcap/bosh/bin/bosh-agent enable-monit-access
+}


rkoster (changes requested):
Why the fallback? the stemcell and the agent are in full control here. So we just need to make sure to these changes go out in the same release as the agent. So please remove the check and the fallback.

rkoster (approved):

rkoster (changes requested):
Actually there should be some more cleanup like in: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/470/files (on my phone because I'm on vacation) but remember there was some setup for the agent itself which should no longer be used. Only jobs should be calling the helper. Also the monit wrapper should be removed.

aramprice (dismissed):

rkoster (dismissed):
Actually: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/470 should just be merged first this PR is just a refinement on the other one.

rkoster (approved):

PR #481, opened 2026-02-27
Use bosh-enable-monit-access if available and fallback to nftables setup

mariash:
Use bosh-enable-monit-access. If it fails then fall back to setting nft rules for vcap user.

This change is required for backwards compatibility with old releases that source the script but don't have exec access to it.

The change also requires https://github.com/cloudfoundry/bosh-agent/pull/416


mariash:
Closing this PR in favor of - https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/482

PR #483, opened 2026-03-02
[ubuntu-noble] Move bosh-enable-monit-access helper to /usr/local/sbin

abg:
Enables access for legacy bosh-releases that setup monit access via the monit-access-helper.sh script and may not have executable access to binaries under /var/vcap/bosh/

Also, adds the "monit-access-helper.sh" pattern from the ubuntu-jammy branch.

### DIFF
--- a/stemcell_builder/stages/bosh_go_agent/apply.sh
+++ b/stemcell_builder/stages/bosh_go_agent/apply.sh
@@ -17,7 +17,7 @@ bosh_agent_version=$(cat ${assets_dir}/bosh-agent-version)
 /usr/bin/meta4 file-download --metalink=${assets_dir}/metalink.meta4 --file=bosh-agent-${bosh_agent_version}-linux-amd64 bosh-agent
 
 mv bosh-agent $chroot/var/vcap/bosh/bin/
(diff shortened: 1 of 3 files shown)


aramprice (approved):
This seems like a reasonable change.

One thought it that if we end up with more bosh-y things maybe it would be worth the hassle (PATH complexities mostly) putting these things into a subdirectory like:
`/usr/local/sbin/bosh/`

... I'm very much on the fence about whether this would be more hassle than its worth.

abg:
Yeah, agreed.   I am not a big fan of dropping things in /usr/local/sbin, but it was following the change to preserve compatibility with older releases on the ubuntu-jammy stemcells.

Based on discussion in [cloudfoundry bosh slack](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1772487431826229?thread_ts=1769638890.623389&cid=C02HPPYQ2) it seems like we want to move towards releases directly calling the `bosh-enable-monit-access` command rather than relying on the `monit-access-helper.sh`.

That command will remain in /var/vcap/bosh/etc so this PR is no longer useful.

PR #487, opened 2026-03-11
Revert "Merge pull request #470 from rkoster/remove-monit-permit-access"

selzoc:
This is actually a breaking change in the stemcell contract, and we found some releases were relying on the previous behavior.  Our intention is to restore this behavior in jammy, and teach the bosh agent to _not_ apply it's nftables rules on jammy.

On noble, the changes made recently remain the same - the bosh agent will be responsible for restricting monit access.

### DIFF
--- a/bosh-stemcell/spec/stemcells/azure_spec.rb
+++ b/bosh-stemcell/spec/stemcells/azure_spec.rb
@@ -40,6 +40,7 @@ describe 'Azure Stemcell', stemcell_image: true do
       its(:content) { should include('"DevicePathResolutionType": "scsi"') }
       its(:content) { should include('"CreatePartitionIfNoEphemeralDisk": true') }
       its(:content) { should include('"PartitionerType": "parted"') }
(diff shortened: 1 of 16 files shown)


aramprice (dismissed):

mariash (dismissed):

rkoster (changes requested):
Let's discuss this in the FI working group meeting

selzoc:
Companion bosh agent change https://github.com/cloudfoundry/bosh-agent/pull/420

aramprice:
If https://github.com/cloudfoundry/bosh-agent/pull/420#discussion_r2924618835 is resolved in favor of using a setting rather than switching on the cgroup version, this PR will need to be updated with the new setting.

selzoc:
Updated the PR to set `UseMonitIptablesFirewall` to `true` so that the bosh agent can use this value to determine to not use the nftables rules on jammy

rkoster (approved):

Issue #485, opened 2026-03-09
Missing ssh keys in /etc/ssh causing issues with ssh


lodener:
In PR #460 `firstboot.sh` is moved to a systemd service. The `ConditionFirstBoot=yes` condition seems to remain unmet, even on first boot across both our Azure and AWS deployments. On Azure `cloud-init` loads in ssh, which seems to trigger ssh key generation:
```
2026-03-09T11:46:49.080407+00:00 7fc2419f-a5a0-462b-64ba-62220c3defed cloud-init[899]: Generating public/private rsa key pair.
2026-03-09T11:46:49.080410+00:00 7fc2419f-a5a0-462b-64ba-62220c3defed cloud-init[899]: Your identification has been saved in /etc/ssh/ssh_host_rsa_key
2026-03-09T11:46:49.080418+00:00 7fc2419f-a5a0-462b-64ba-62220c3defed cloud-init[899]: Your public key has been saved in /etc/ssh/ssh_host_rsa_key.pub
...etc
```
On AWS this does not happen, and no ssh keys are generated:
```
#  ls -l /etc/ssh/
total 648
-rw-r--r--. 1 root root 620042 Aug 26  2025 moduli
-rw-r--r--. 1 root root   1649 Aug 26  2025 ssh_config
drwxr-xr-x. 2 root root   4096 Aug 26  2025 ssh_config.d
-rw-------. 1 root root   4076 Mar  8 08:09 sshd_config
drwxr-xr-x. 2 root root   4096 Mar  9 10:31 sshd_config.d
```
This in turn causes `ssh.service` to fail:
```
/var/log/syslog:2026-03-09T10:40:54.577316+00:00 8cf11777-b925-445a-42d0-32e893894179 sshd[41503]: Unable to load host key: /etc/ssh/ssh_host_rsa_key
/var/log/syslog:2026-03-09T10:40:54.577708+00:00 8cf11777-b925-445a-42d0-32e893894179 sshd[41503]: Unable to load host key: /etc/ssh/ssh_host_ecdsa_key
/var/log/syslog:2026-03-09T10:40:54.577962+00:00 8cf11777-b925-445a-42d0-32e893894179 sshd[41503]: Unable to load host key: /etc/ssh/ssh_host_ed25519_key
[...]

PR #489, opened 2026-03-14
Restore "Fix race condition causing sshd start failure during provisioning"

s4heid:
Following the revert in #488, this pull request restores #460, but using the previous file-based method for creating and checking the `firstboot_done` file. It should be compatible with AWS and help address the issue noted in #485. However, I currently don't have access to an AWS environment to verify this.

On Azure:

```
systemctl status firstboot.service
○ firstboot.service - Run first boot tasks
     Loaded: loaded (/etc/systemd/system/firstboot.service; enabled; preset: enabled)
     Active: inactive (dead) since Fri 2026-03-13 23:36:30 UTC; 6min ago
   Main PID: 1006 (code=exited, status=0/SUCCESS)
        CPU: 1.086s
```


metskem:
We created a new stemcell from your branch, repaved a VM with it, and we can confirm the problem is fixed.
Thanks!

rkoster (approved):

PR #499, opened 2026-03-27
Remove redundant be_symlink assertion

rkoster:
Remove redundant be_symlink assertion in warden stemcell spec since be_linked_to already asserts the symlink target and ShelloutTypes::File does not implement #symlink?. This avoids calling a missing predicate and fixes CI failures.\n\nIf we prefer the more explicit assertion, we can instead implement ShelloutTypes::File#symlink? in a follow-up change.


rkoster:
@coderabbitai review

aramprice (approved):

aramprice:
Thanks!

Issue #504, opened 2026-03-31
[warden/boshlite] Missing SSH keys in /etc/ssh causing issues with ssh


mike-jc:
## Service `firstboot` is not enabled, and SSH keys in /etc/ssh are missing

This is somewhat related to [issue 485](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/485). But as I understand it, it is fixed and works with hyperscalers (at least it's been tested on AWS). 

I assume this is something specific to the Bosh Lite stemcell type.

### Setup

**Stemcell**: Ubuntu Noble v1.305 for BOSH LIte (bosh-warden-boshlite-ubuntu-noble)
**CPI**: Docker CPI
**Systemd**: 255.4-1ubuntu8.14
**systemd-detect-virt**: docker

```
# ls -la /proc/1/exe
lrwxrwxrwx 1 root root 0 Mar 27 18:56 /proc/1/exe -> /usr/lib/systemd/systemd

# systemd-detect-virt
docker

# systemctl is-system-running
running
```

### Steps to reproduce

1. Create a Bosh environment with Bosh Lite and Docker CPI.
2. Deploy any CloudFoundry component with bosh-warden-boshlite-ubuntu-noble stemcell.
3. Try to SSH to the created VM and see the error.

### Symptoms

VM is not accessible via SSH after creation. When I tried to SSH to it, I got the error:
```
# bosh -d haproxy1 ssh haproxy/0
...
Task 11 | 18:58:23 | Error: Action Failed ssh: Getting host public key: Unable to read host public key file: /etc/ssh/ssh_host_rsa_key.pub: Opening file /etc/ssh/ssh_host_rsa_key.pub: open /etc/ssh/ssh_host_rsa_key.pub: no such file or directory
```

**Failure chain**
 
- The firstboot service is disabled for an unknown reason, although it has been enabled during the OS image build.
- The disabled firstboot service has not been started, and the marker file firstboot_done has not been created.
- The SSH keys have not been created.
- The SSH service could not be started since it waits for the firstboot service.

```
# systemctl show firstboot.service | grep -E "(LoadState|ActiveState|UnitFileState)"
LoadState=loaded
ActiveState=inactive
UnitFileState=disabled

# ll /root/firstboot_done
ls: cannot access '/root/firstboot_done': No such file or directory

# journalctl -u ssh.service -u sshd.service
[...]

PR #486, opened 2026-03-11
Add Rosetta x86_64 emulation compatibility for warden stemcells

rkoster:
## Summary

- Add systemd drop-in overrides to disable security features that conflict with Rosetta's JIT compilation on Apple Silicon Macs
- Mask `systemd-binfmt.service` which fails under Rosetta emulation
- Add tests to verify the Rosetta compatibility configurations

## Background

When running warden stemcells under Rosetta x86_64 emulation on Apple Silicon Macs, several systemd services fail because their security hardening features (`MemoryDenyWriteExecute`, `SystemCallFilter`, etc.) conflict with Rosetta's JIT compilation which requires writable+executable memory.

This change is based on the patches from [instant-bosh](https://github.com/rkoster/instant-bosh/blob/32c294a319522bbf633cb5f3092500df21aee6a1/apple-silicon/stemcell/Dockerfile), converted into proper stemcell builder fixes that only affect the warden stemcell.

### New file: `stemcell_builder/stages/base_warden/assets/rosetta-compat.conf`

Systemd drop-in configuration that disables security features:
- `MemoryDenyWriteExecute=no`
- `SystemCallArchitectures=` (cleared)
- `SystemCallFilter=` (cleared)
- `LockPersonality=no`
- `NoNewPrivileges=no`

### Modified: `stemcell_builder/stages/base_warden/apply.sh`

Installs the drop-in configuration for the following services:
- `systemd-journald`
- `systemd-resolved`
- `systemd-networkd`
- `systemd-logind`
- `systemd-timesyncd`
- `auditd`

Also masks `systemd-binfmt.service`.

### Modified: `bosh-stemcell/spec/stemcells/warden_spec.rb`

Added tests to verify:
- Drop-in configuration files exist for all affected services
- Configuration contains expected directives
- `systemd-binfmt.service` is masked (symlinked to `/dev/null`)

PR #506, opened 2026-04-01
Ubuntu Noble warden-specific systemd hardening

mike-jc:
**This is for Ubuntu Noble stemcell.**

These changes are related to issues:
- [warden/boshlite: Missing SSH keys in /etc/ssh causing issues with ssh](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/504)
- [Necessary systemd services could be removed during hardening](https://github.com/cloudfoundry/bosh-docker-cpi-release/issues/58)

Added a new stage `base_systemd_clean` to the `warden_stages` collection, which is applied during stemcell build, and removes all custom and optional systemd services, keeping only the necessary minimum set.


mkocher:
Given that we use our containerized stemcells for a lot of testing, I'd be inclined to have it mimic production as much as conveniently possible. I'd prefer to see this an explicit list of services to disable in a container instead of an allow list of services. We recently did some of this in #486 (noble) and #500 (jammy).

mike-jc:
> Given that we use our containerized stemcells for a lot of testing, I'd be inclined to have it mimic production as much as conveniently possible. I'd prefer to see this an explicit list of services to disable in a container instead of an allow list of services. We recently did some of this in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/486 (noble) and https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/500 (jammy).

Thank you @mkocher, that's a good point.

I removed the allowlisting of systemd services and tested the mask list approach. The {{systemd-binfmt}} service is already masked, and testing confirms this is sufficient for warden/boshlite on both Jammy and Noble: during testing with the latest official Noble stemcell and the Jammy stemcell built on [PR 500](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/500), along with a dev release containing [PR 60](https://github.com/cloudfoundry/bosh-docker-cpi-release/pull/60) for Docker CPI, all HAProxy acceptance tests passed successfully.

So I close this PR and the PR for Jammy. We need only Docker CPI changes to have stemcells work there.

PR #507, opened 2026-04-01
Ubuntu Jammy warden-specific systemd hardening

mike-jc:
**This is for Ubuntu Jammy stemcell.**

These changes are related to issues:
- [warden/boshlite: Missing SSH keys in /etc/ssh causing issues with ssh](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/504)
- [Necessary systemd services could be removed during hardening](https://github.com/cloudfoundry/bosh-docker-cpi-release/issues/58)

Added a new stage `base_systemd_clean` to the `warden_stages` collection, which is applied during stemcell build, and removes all custom and optional systemd services, keeping only the necessary minimum set.


mike-jc:
Closing this PR as we have decided not to proceed with the allow list approach in favor of the mask list approach, see [the comment in PR 506](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/506#issuecomment-4177604034).

PR #508, opened 2026-04-02
Allow building a stemcell with a custom agent (jammy)

mkocher:
Set BOSH_AGENT_BIN_PATH=./bosh-agent to provide a bosh agent binary to be injected into the stemcell being built. Useful when doing agent development work.

I don't love environment variables, but since this is only a development flow it doesn't seem like it makes much sense as a rake task argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ The arguments to `stemcell:build_with_local_os_image` are:
    of the most recent release and add one, e.g.: "0.0.7" → "0.0.8". If not
    specified, it will default to "0000".
 
+While building a stemcell for development purposes it is possible to set the BOSH_AGENT_BIN_PATH environment variable to inject an agent binary into the stemcell instead of downloading one.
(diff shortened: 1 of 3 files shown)


aramprice (approved):

rkoster (changes requested):
Don't like how intrusive these changes are. For iterating on an agent change I would highly recommend using the repack helper scripts here: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/tree/ubuntu-jammy/scripts/repack-helpers

When working with the docker cpi a simple Dockerfile and the lite stemcell support (which just need a reference to a local image) also work really well.

beyhan:
@mkocher Closing because of inactivity. Please reopen if this is still needed.

aramprice:
Discussed in the FIWG meeting, "repack" already handles this, and is likely faster.

Issue #450, opened 2025-10-23
Merge forward from Jammy -> Noble


aramprice:
See recent Azure network interface changes

PR #510, opened 2026-04-02
Merge jammy into noble

aramprice:
See:
- https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/450
- https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/509


aramprice:
Will bypass Easy CLI: all commits were previously accepted in https://github.com/cloudfoundry/bosh-stemcells-ci

See: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/509#issuecomment-4178535511

PR #514, opened 2026-04-03
Merge forward from ubuntu-jammy

aramprice:


aramprice:
Will bypass Easy CLI: all commits were previously accepted in https://github.com/cloudfoundry/bosh-stemcells-ci

PR #531, opened 2026-04-09
CI: add bump-deps task for gems

aramprice:
- update ruby-version to 3.3; version 3.2 is out of support
- add github action to run bosh-stemcell/ specs
- add dependabot config to update github actions


ci/pipelines/builder.yml
```diff
@@ -729,6 +730,39 @@ jobs:
...
+        output-repo: bosh-linux-stemcell-builder
+      params:
+        GIT_USER_NAME: CI Bot
+        GIT_USER_EMAIL: bots@cloudfoundry.org
+        PACKAGE: ruby-((.:ruby_version))
+        VENDOR: true
+    - task: test-unit
+      file: bosh-stemcells-ci/ci/tasks/test-unit.yml
+      image: bosh-integration-image
+      privileged: true
```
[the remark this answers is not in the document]
aramprice: fixed

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+---
+version: 2
```
Alphasite, line 2: Do we have both dependabot and bump deps?

Alphasite (approved):
(An aside: It would be cool if we pinned the apt package versions as well. But I guess the current approach is fine too)

.github/dependabot.yml
aramprice: Dependabot is to update Github actions

aramprice:
> (An aside: It would be cool if we pinned the apt package versions as well. But I guess the current approach is fine too)

This is sort a happening with the build-time marker file so that an apt repo from a given (¿hour, day?) can be used to more or less recreate a given build.

Issue #584, opened 2026-04-30
ubuntu-noble warden stemcell: BOSH DNS runtime config needs recursor adjustment for BOSH Lite on GCP


hamideh-amini-khezrabad:
When using the ubuntu-noble warden stemcell on BOSH Lite (GCP), DNS resolution breaks because the current BOSH DNS runtime config sets "disable_recursors: true" for ubuntu-noble.

A similar dedicated DNS runtime config adjustment needs to be added for the warden stemcell with:

disable_recursors: false
recursors:
  - 169.254.169.254
  
169.254.169.254 is GCP's internal metadata server required for DNS resolution on BOSH Lite.

See also: [2b7f09f](https://github.com/cloudfoundry/bosh-deployment/commit/2b7f09f2cf5de25819a9fdab4bcda60a66396446)

aramprice:
I asked Gemini to weigh in on this and got the following response (I have NOT validated this):

---

To enable access to the GCP metadata server (`169.254.169.254`) on Ubuntu systems, two primary `systemd` components work together:

1. **`systemd-networkd.service`**: This is the core unit that enables *network routing* to the metadata IP. On GCP, the route to the metadata server is provided by the DHCP server via **DHCP Option 121** (classless static routes). `systemd-networkd` must be running and configured to accept DHCP routes (`UseRoutes=true`) to install the `169.254.169.254/32` route into the kernel routing table.
2. **`google-guest-agent.service`** (and historically `google-network-daemon.service`): Provided by the `google-guest-agent` package, this systemd unit actively manages the guest environment by polling `169.254.169.254` for metadata changes, setting up OS Login, and ensuring internal DNS routing (via `gce-resolved.conf`) is pointed at the metadata server.

If instances deployed from your stemcell are timing out when attempting to reach the metadata server, it is usually because `systemd-networkd` is misconfigured to ignore the DHCP host route, or the GCP guest agent services are masked/disabled.

### How to update `bosh-linux-stemcell-builder`

To ensure these units are present, installed, and enabled on GCP stemcells, you must modify the build stages in the `cloudfoundry/bosh-linux-stemcell-builder` repository. The stemcell builder uses "stages" (shell scripts that run inside a `chroot` of the OS image) to configure the OS.

Here is how you would update the repository:

#### 1. Update/Verify the GCP Package Stage
[...]

PR #608, opened 2026-06-03
add ops file for noble warden, use it in deploy-director script

dudejas:
addresses: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/584


neddp (approved):

neddp:
Potential fix for the root cause: https://github.com/cloudfoundry/bosh-dns-release/pull/156

PR #610, opened 2026-06-04
CI: Remove "bats" string from stemcell directors

selzoc:
The bats cleanup job in the director pipeline uses leftovers with a filter "bats", and can sometimes delete stemcell vms that have that string.  Instead, use "test-stemcells-OS" as the tag.

Also, remove unused VARS_TAG parameter.

See also https://github.com/cloudfoundry/bosh/pull/2742

### DIFF
--- a/ci/pipelines/builder.yml
+++ b/ci/pipelines/builder.yml
@@ -4,6 +4,7 @@ anchors:
     name: &ci_bot_name CI Bot
 
 #@ load("@ytt:data", "data")
+#@ bats_director_tag = "test-stemcells-" + data.values.stemcell_details.os_short_name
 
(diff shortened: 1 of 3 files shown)


ystros (approved):

selzoc:
Ran
```
gcloud compute firewall-rules update stemcell-builder-integration-22 \
  --target-tags=bat,test-stemcells-jammy \
  --project=cloud-foundry-310819
```
to update the firewall rules, then https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/ubuntu-jammy-builder/jobs/ensure-integration-network/builds/2 to verify.

PR #620, opened 2026-06-09
Remove CentOS-specific code and assets

neddp:
CentOS stemcells are no longer built, so this drops dead stages, assets, and code paths. Also removes the default_su_directive override (already removed on ubuntu-noble), which silently overrode the correct 'su root syslog' set globally in ubuntu-logrotate.conf.


aramprice (approved):
Looks good, thank you for tackling this!

 Wondering if we should also remove this test for `rpm`, or is it relevant because of `openSUSE`?

```ruby
  describe package("rpm") do
    it { should_not be_installed }
  end
```
^ https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/remove-centos-cleanup-jammy/bosh-stemcell/spec/os_image/ubuntu_spec.rb#L13-L15

bosh-stemcell/lib/shellout_types/service.rb
```diff
@@ -27,7 +27,7 @@ def check_service_enabled(runlevel)
       stdout, stderr, status = @chroot.run("cat", "/etc/*release")
       raise stderr.to_s if status != 0
 
-      raise "Cannot determine Linux distribution: #{stdout}" unless /Ubuntu|CentOS|openSUSE/.match?(stdout)
+      raise "Cannot determine Linux distribution: #{stdout}" unless /Ubuntu|openSUSE/.match?(stdout)
```
aramprice, line 30: Is `openSUSE` still supported?

neddp:
Hi @aramprice,

This is something I planned to bring up on the community meeting. If we confirm that SUSE is also not being used, we can do the cleanup separately.

I would like to go over the other repositories as well and do the same, especially the docs-bosh. Where is the correct place to open an issue to track the clean up?

aramprice:
Maybe an umbrella issue would be best in the community repo? I'm not certain though.

Issue #637, opened 2026-06-12
PR #599 regresses warden/Docker stemcells: broken awk pattern and renamed function breaks restrict-monit-api-access


colins:
@gberche-orange @rkoster 

## Summary

PR #599 (merged Jun 5 2026, fixing #585) introduced two regressions that prevent the BOSH agent from bootstrapping on **warden/bosh-lite stemcells** (confirmed broken in stemcell 1.1234).

---

## Regression 1 — broken awk pattern (`1e4a11482`)

The fix changed the cgroup2 mount detection from:

```bash
cgroup_mount="$(awk '$3 == "cgroup2" { print $2 }' /proc/self/mounts)"
```

to:

```bash
cgroup_mount="$(awk '$1 == "cgroup2" && $3 == "cgroup2" { print $2 }' /proc/self/mounts)"
```

The intent was to avoid matching extra cgroup2 mounts added by Cilium workloads (#585). However, this breaks **warden/Docker** containers where the device field (`$1`) for the cgroup2 mount is `"cgroup"`, not `"cgroup2"`:

```
cgroup  /sys/fs/cgroup  cgroup2  rw,...   ← $1 == "cgroup", never "cgroup2"
```

Result: `cgroup_mount` is always empty → `nb_matching_cgroup_mounts=0` → `permit_monit_access` returns 1 → both the agent and monit runit run scripts exit immediately (they source this helper with `set -e`) → **BOSH agent never starts on any warden VM**.

**The correct fix for #585** is to use `exit` after the first match, which selects the canonical system cgroup2 mount regardless of device name, and naturally ignores any extra Cilium mounts (which appear later in `/proc/self/mounts`):

```bash
cgroup_mount="$(awk '$3 == "cgroup2" { print $2; exit }' /proc/self/mounts)"
```

---

## Regression 2 — stale function name in `restrict-monit-api-access` (`50cc4d1fc`)

Commit `50cc4d1fc` renamed `monit_using_unified_cgroup_v2` → `system_using_unified_cgroup_v2` in `monit-access-helper.sh` but did **not** update `restrict-monit-api-access`, which still calls the old name:

```bash
# restrict-monit-api-access (unchanged by PR #599)
source /var/vcap/bosh/etc/monit-access-helper.sh
if monit_using_unified_cgroup_v2; then   ← function no longer exists
```

[...]

PR #599, opened 2026-05-13
Harden monit-access-helper.sh cgroupv2 mount point detection

gberche-orange:
Restrict the inspection of /proc/self/mounts to cgroupv2 device (1st column) in addition to existing cgroup fstype (column 3).

Also fail fast in case of multiple detected mount points.

Fix #585

----

NOTE: this repository uses a "Merge Forward" strategy

Changes should be made in the earliest applicable branch, and
merged forward through subsequent branches.
1. Create a PR into the oldest branch (`ubuntu-<short_name>`)
2. After this PR has been merged create a `merge-to-<next_short_name>` branch
3. Merge `ubuntu-<short_name>` into `merge-to-<next_short_name>`
4. Create a PR to merge `merge-to-<next_short_name>` into `ubuntu-<next_short_name>`
5. Repeat as needed for subsequent branches

PR #640, opened 2026-06-12
Fix issue 637

gberche-orange:
Reverts #599 with #638, and harden cgroupv2 mounts detection by selecting the first mount

Hope this helps. Note that I'm unable to do further tests, and rely on suggested fix by @colins in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/637

> This was root-caused and verified by manually patching the three issues in a running stemcell 1.1234 container, after which the BOSH agent bootstrapped successfully.
----

1e4a11482 was filtering on the cgroupv2 device to prevent a `cgroup_mount` variable with multiline content failing downsteam without clear errors.

cilium-originating mount
```
cat /proc/self/mounts | grep cgroup2
     # device  mount_point         fs_type dummy
    cgroup2 /sys/fs/cgroup/unified cgroup2 rw,nosuid,nodev,noexec,relatime 0 0
    none    /run/cilium/cgroupv2   cgroup2 rw,relatime 0 0
```

however, in warden/docker stemcells, the device is cgroup, which introduced regression #637
```
cat /proc/self/mounts | grep cgroup2
     # device  mount_point         fs_type dummy
     cgroup  /sys/fs/cgroup        cgroup2 rw,...
```

Applying suggestion by @colins in https://github.com/cloudfoundry/bosh-linux-stemcell-builder/issues/637 to instead rely on the chronological ordering of mount points, and select the canonical cgroup2 mount point first added by the kernel during boot process.

https://man7.org/linux/man-pages/man5/proc_pid_mounts.5.html
> /proc/self/mounts, lists the mounts of the process's own mount namespace.  The format of this file is documented in [fstab(5)](https://man7.org/linux/man-pages/man5/fstab.5.html).

https://man7.org/linux/man-pages/man5/fstab.5.html
> The order of records in fstab is important because [fsck(8)](https://man7.org/linux/man-pages/man8/fsck.8.html), [mount(8)](https://man7.org/linux/man-pages/man8/mount.8.html), and [umount(8)](https://man7.org/linux/man-pages/man8/umount.8.html)
>        sequentially iterate through fstab doing their thing

https://man7.org/linux/man-pages/man7/cgroups.7.html
>   Note that on many modern systems, systemd(1) automatically mounts
>        the cgroup2 filesystem at /sys/fs/cgroup/unified during the boot
>        process.

### DIFF
--- a/stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh
+++ b/stemcell_builder/stages/bosh_monit/assets/monit-access-helper.sh
@@ -29,7 +29,7 @@ permit_monit_access() {
         # cgroupv2 (unified hierarchy)
         # Create a sub-cgroup under the current process's cgroup and move into it.
         # The iptables rules match on this cgroup path.
-        cgroup_mount="$(awk '$3 == "cgroup2" { print $2 }' /proc/self/mounts)"
+        cgroup_mount="$(awk '$3 == "cgroup2" { print $2; exit }' /proc/self/mounts)"
         current_cgroup="$(grep '^0::' /proc/self/cgroup | cut -d: -f3)"
         if [ -z "${cgroup_mount}" ] || [ -z "${current_cgroup}" ]; then
             echo "permit_monit_access: unable to resolve cgroup v2 mount or path" >&2


beyhan:
FYI: @gberche-orange we look into prs after the AI feedback is resolved.

gberche-orange:
@beyhan thanks for the heads up. I've now rebased the PR to clear unrelated outdated bot reviews, and addressed latest bot review.

aramprice (approved):

PR #644, opened 2026-06-12
fix: replace files dropped in merge forward

aramprice:
Fixes:
```
  1) AWS Stemcell installed by bosh_aws_agent_settings /var/vcap/bosh/agent.json sets InstanceStorageDevicePattern for NVMe instance storage
     Failure/Error: expect(config.dig("Platform", "Linux", "InstanceStorageDevicePattern")).to eq("/dev/nvme*n1")
     
       expected: "/dev/nvme*n1"
            got: nil
     
       (compared using ==)
     # ./spec/stemcells/aws_spec.rb:33:in `block (4 levels) in <top (required)>'

  2) AWS Stemcell installed by bosh_aws_agent_settings /var/vcap/bosh/agent.json sets InstanceStorageManagedVolumePattern to exclude EBS volumes
     Failure/Error: expect(config.dig("Platform", "Linux", "InstanceStorageManagedVolumePattern")).to eq("/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*")
     
       expected: "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*"
            got: nil
     
       (compared using ==)
     # ./spec/stemcells/aws_spec.rb:38:in `block (4 levels) in <top (required)>'
```
^ https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/ubuntu-noble-builder/jobs/build-aws-xen-hvm/builds/631#L6a2a20e2:15435:15451

### DIFF
--- a/stemcell_builder/stages/bosh_aws_agent_settings/assets/agent.json
+++ b/stemcell_builder/stages/bosh_aws_agent_settings/assets/agent.json
@@ -6,7 +6,10 @@
       "CreatePartitionIfNoEphemeralDisk": true,
       "ServiceManager": "systemd",
       "DiskIDTransformPattern": "^vol-(.+)$",
-      "DiskIDTransformReplacement": "nvme-Amazon_Elastic_Block_Store_vol${1}"
+      "DiskIDTransformReplacement": "nvme-Amazon_Elastic_Block_Store_vol${1}",
+      "UseMonitIptablesFirewall": true,
+      "InstanceStorageDevicePattern": "/dev/nvme*n1",
+      "InstanceStorageManagedVolumePattern": "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*"
     }
   },
   "Infrastructure": {


aramprice:
@Ivaylogi98 or @neddp would one of you mind weighing in on whether this is the right fix for the errors in CI from my recent merge forward Jammy => Noble? Thanks!

stemcell_builder/stages/bosh_aws_agent_settings/assets/agent.json
aramprice, line 1: Should `stemcell_builder/stages/bosh_aws_agent_settings/assets/agent.md` be removed as well?

neddp:
Hi @aramprice,

Thanks for looking into this. I noticed that the same applies for all providers and `cp` with the `agent.json` is being used on noble, while the heredoc is used across jammy. I don't understand how git handled this, but it looks like the `agent.json` remains outdated this way.
The assets directory is completely removed on jammy, for the heredoc approach.

I think we should do the same for the other provider's agent.sh files on noble to keep it consistent in the future.

aramprice:
@neddp - I switched (back) to the `cp` style we have in noble. 

Would you mind taking a look to validate that the AWS `agent.json` is correct?

Feel free to merge if it looks good. I'm OOO until 2026-06-16

Thanks!

mkocher (approved):
This is breaking the Resolute build too, and the fix looks good to me. I'll go ahead and merge this.

neddp:
Opened a PR that matches the noble structure on jammy - https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/646

PR #646, opened 2026-06-14
refactor: align jammy agent settings to cp+assets approach

neddp:
Replace heredoc-based `apply.sh` scripts with static `assets/agent.json` files (the same approach already used on ubuntu-noble). This makes the two branches structurally identical so future jammy->noble merge forwards cannot silently resolve `apply.sh` conflicts in favor of the wrong version.
As seen here: #644 

- Delete `prelude_agent.bash` (PartitionerType: parted is now inlined as a static field in each `assets/agent.json`, as the function has unconditionally returned that string since 59a9c29c7 in 2019)
- Replace all 8 `bosh_*_agent_settings/apply.sh` heredocs with a single `cp $assets_dir/agent.json` line
- Add `assets/agent.json` for every stage carrying the full config (including `UseMonitIptablesFirewall` added in 8104abdd1)

### DIFF
--- a/stemcell_builder/lib/prelude_agent.bash
+++ b/stemcell_builder/lib/prelude_agent.bash
@@ -1,4 +1,0 @@
-
-function get_partitioner_type_mapping {
-  echo '"PartitionerType": "parted",'
-}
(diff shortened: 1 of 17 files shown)


aramprice (approved):
Nice. Thank you for back-porting this change!

PR #648, opened 2026-06-18
Fix test stemcells

mkocher:
Assorted fixes to make test-stemcells pass for the first time:
* add bpm to acceptance tests manifest since syslog requires it on Resolute
* fix some test cleanup code
* fix some pam spec failures


aramprice (dismissed):

aramprice:
backported to jammy

PR #657, opened 2026-06-19
Restore non-essential systemd unit cleanup for jammy warden

selzoc:
[bosh-docker-cpi-release#60](https://github.com/cloudfoundry/bosh-docker-cpi-release/pull/60) moved the cleanup of non-essential systemd units out of the Docker CPI and into the stemcell, on the basis that the stemcell would mask anything that can't run in a container (resolving [bosh-docker-cpi-release#58](https://github.com/cloudfoundry/bosh-docker-cpi-release/issues/58) and #504). On the jammy line, only the systemd-binfmt mask landed (#500); the rest of the CPI's allow-list was never reproduced here.

As a result, warden stemcells started with exec /sbin/init now boot the full stock systemd unit set. In a BOSH director the stock units contend with the monit-managed bpm jobs and the director fails to converge (e.g. the postgres role is never created). This regresses any consumer running a Docker-CPI director on a jammy warden stemcell ≥ the build that pairs with Docker CPI ≥ 0.2.9.

This restores the CPI's historical allow-list in base_warden, removing the non-essential .wants entries at build time. It uses -delete (≡ systemctl disable) rather than mask, matching the CPI's original semantics: units are dropped from the boot sequence but can still start as dependencies of a kept unit. runit, ssh, dbus, journald, logrotate, systemd-tmpfiles, systemd-user-sessions, and the bosh-agent are preserved. Deriving the set at build time keeps it correct as the package set changes.

Verified against ubuntu-jammy-stemcell:1.1250: 13 essential .wants entries kept, ~75 non-essential removed. This is the same prune Docker CPI 0.2.3 applied, which is green on cgroups-v1 workers. Adds a warden_spec assertion that non-essential units are removed and essential ones remain.

### DIFF
--- a/bosh-stemcell/spec/stemcells/warden_spec.rb
+++ b/bosh-stemcell/spec/stemcells/warden_spec.rb
@@ -58,4 +58,34 @@ describe "Warden Stemcell", stemcell_image: true do
       its(:content) { should include('"UseMonitIptablesFirewall": true') }
     end
   end
+
+  context "systemd unit cleanup for warden containers" do
+    # The Docker CPI runs warden stemcells with `exec /sbin/init`. base_warden
+    # strips non-essential stock systemd units from the boot sequence so they
+    # don't contend with the monit-managed bpm jobs in the BOSH director
+    # container (symptom: postgres role never created, bosh/0 never converges).
+    # Keep-list mirrors the historical Docker CPI allow-list. See
+    # base_warden/apply.sh and cloudfoundry/bosh-docker-cpi-release#60.
+    keep_patterns = %w[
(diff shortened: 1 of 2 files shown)


aramprice (approved):
Seems reasonable though I'm not 100% confident in my assessment of systemd unit requirements 😬

mkocher (changes requested):
I'm in favor of cleaning up units that error in a container, but I think we should do that surgically instead of relying on an allow list

selzoc:
Yes I plan to dig deeper and come back when I have some time to basically make the inverse of this - examine a running container for failed units and denylist them.

PR #677, opened 2026-06-29
Fix smoke suite: upload bpm-release, drop unused os-conf

neddp:
The smoke manifest declares bpm as a required release (added in ba67a704d), but the BeforeSuite was missed in that commit - only the parallel syslogrelease/smoke_suite_test.go was updated. This caused 'Release bpm doesn't exist' during deploy.

While here, drop os-conf upload: the smoke manifest never references os-conf, so uploading it added a redundant compile/upload step.

This should unblock the test-stemcells-ipv4 pipeline.


ramonskie (approved):

aramprice (approved):

mkocher (approved):

mkocher (approved):

mkocher:
Apologies, I fixed this on the Resolute branch in f8821341c4a9b1fffeec51af4d535e06198233c0, should have done that on Jammy and merged it forward.

PR #692, opened 2026-07-16
Remove legacy Azure ifupdown configuration

s4heid:
The Azure cloud-init datasource bootstraps access to IMDS and renders the reported NIC configuration through netplan to systemd-networkd. The eth0 DHCP entry in `/etc/network/interfaces` is therefore redundant.

On Noble, ifupdown starts dhcpcd (because dhclient is not installed). This leaves dhcpcd and systemd-networkd managing eth0 concurrently. dhcpcd occupies the DHCPv6 client socket on UDP port 546, causing networkd to fail with "Address already in use" and preventing IPv6 assignment - I discovered this when testing IPv6 related changes on Azure: https://github.com/cloudfoundry/bosh-azure-cpi-release/pull/746.

Therefore, remove the legacy config and its corresponding stemcell test so Azure networking has a single owner.

### DIFF
--- a/bosh-stemcell/spec/stemcells/ubuntu_spec.rb
+++ b/bosh-stemcell/spec/stemcells/ubuntu_spec.rb
@@ -146,22 +146,6 @@ describe "Ubuntu 24.04 stemcell image", stemcell_image: true do
     end
   end
 
-  context "installed by system-azure-network", {
-    exclude_on_alicloud: true,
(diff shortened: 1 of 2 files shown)


naioja:
+1 LGTM

HappyTobi (commented):
+1 LGTM

neddp (approved):

PR #694, opened 2026-07-24
fix: enable systemd forwarding to syslog

mkocher:
Make systemd forward logs to syslog. This is breaking the bosh-agent tests which validate that logs flow through syslog to /var/log/bosh-agent.log.


ramonskie:
why is this merged by the same person who created the pr?

and second what is this pr about.
as we already logged and symlinked to the old place.
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-noble/stemcell_builder/stages/bosh_go_agent/apply.sh#L53

and we already had a syslog identifier.
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-noble/stemcell_builder/stages/bosh_go_agent/assets/bosh-agent.service#L16

PR #696, opened 2026-07-27
ci: Handle concurrent deletions in CI

ystros:
This pipeline exists in the ubuntu-jammy, ubuntu-noble, and ubuntu-resolute
pipelines. They all run at the same time. The cleanup of 3 year old images is
looking at ALL distributions in each pipeline. This caused failures where one
pipeline would delete an AMI and then the other 2 pipelines would try to delete
that AMI again and fail.

We cannot limit the cleanup to the current pipeline's OS, because we want to
clean up old AMIs even after a stemcell line is EOL and its pipeline has been
deleted. So this commit makes the cleanup script tolerant to the AMI having
been deleted.

This also removes an unused snapshot_id cleanup mechanism.

### DIFF
--- a/ci/tasks/light-aws/cleanup-ami.sh
+++ b/ci/tasks/light-aws/cleanup-ami.sh
@@ -59,16 +59,6 @@ for region in ${ami_destinations}; do
     ami_list=$(jq -s '.[0] + .[1]' <(echo "${ami_list}") <(echo "${results}"))
   fi
 
-  if [ -n "${snapshot_id:-}" ]; then
-    results=$(aws ec2 describe-images \
-            --owners self \
-            --output json \
-            --region "${region}" \
-            --filters "Name=block-device-mapping.snapshot-id,Values=${snapshot_id}" \
-            --query "${past_due_query}" | jq 'reverse | del(.[range(env.ami_keep_latest|tonumber)])')
-    ami_list=$(jq -s '.[0] + .[1]' <(echo "${ami_list}") <(echo "${results}"))
-  fi
-
   # 'ami_list' is a json array of objects, each object is an ami and its snapshot
   for row in $(echo "${ami_list}" | jq -r '.[] | @base64'); do
     _jq() {
@@ -86,11 +76,5 @@ for region in ${ami_destinations}; do
     aws ec2 deregister-image \
       --image-id "$(_jq '.ImageId')" \
       --region "${region}"
-
-    if [ "${snapshot_id:-}" != "$(_jq '.SnapshotId')" ]; then
-      aws ec2 delete-snapshot \
-        --snapshot-id "$(_jq '.SnapshotId')" \
-        --region "${region}"
-    fi
   done
 done
--- a/ci/tasks/light-aws/cleanup-ami.sh
+++ b/ci/tasks/light-aws/cleanup-ami.sh
@@ -30,6 +30,28 @@ if [ -n "${ami_role_arn:-}" ]; then
   export AWS_PROFILE=resource_account
 fi
 
+# Every stemcell pipeline runs the published sweep against the same account, so two
+# pipelines can select the same AMI before either has deleted it. This handles
+# expected potential errors that could occur on a collision.
+run_tolerating_missing() {
+  local output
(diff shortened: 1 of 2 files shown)


ci/tasks/light-aws/cleanup-ami.sh
```diff
@@ -30,6 +30,28 @@ if [ -n "${ami_role_arn:-}" ]; then
   export AWS_PROFILE=resource_account
 fi
 
+# Every stemcell pipeline runs the published sweep against the same account, so two
+# pipelines can select the same AMI before either has deleted it. This handles
+# expected potential errors that could occur on a collision.
+delete_tolerating_missing() {
+  local output
+  if output="$("$@" 2>&1)"; then
+    [ -n "${output}" ] && echo "${output}"
```
Alphasite: nit: echo should probably be printf

ci/tasks/light-aws/cleanup-ami.sh
```diff
@@ -30,6 +30,28 @@ if [ -n "${ami_role_arn:-}" ]; then
...
+    return 0
+  fi
+
+  case "${output}" in
+    *InvalidAMIID.Unavailable*|*InvalidAMIID.NotFound*|*InvalidSnapshot.NotFound*)
```
Alphasite: gemini seems to think there is 1 more condition which may be worth a looksie at:

> InvalidSnapshot.InUse Race Condition
> When multiple pipelines are sweeping AMIs, they usually follow two steps:
> 
>    1. Deregister the AMI.
>    1. Delete the underlying EBS Snapshots.
> 
> Because AWS is eventually consistent, if Pipeline A deregisters the AMI, AWS takes a few moments to "unlock" the underlying snapshot. If Pipeline B hits the snapshot deletion step during this window, AWS will throw InvalidSnapshot.InUse (not NotFound). If this function doesn't tolerate InUse in a concurrent environment, Pipeline B will fail completely.

ci/tasks/light-aws/cleanup-ami.sh
```diff
@@ -59,16 +81,6 @@ for region in ${ami_destinations}; do
     ami_list=$(jq -s '.[0] + .[1]' <(echo "${ami_list}") <(echo "${results}"))
   fi
 
-  if [ -n "${snapshot_id:-}" ]; then
```
Alphasite, line 62: q: im guessing this isn't used elsewhere in the pipelines? id worry about leaking jobs in other steps where this was being misused as a dual purpose script.

ci/tasks/light-aws/cleanup-ami.sh
```diff
@@ -30,6 +30,28 @@ if [ -n "${ami_role_arn:-}" ]; then
   export AWS_PROFILE=resource_account
 fi
 
+# Every stemcell pipeline runs the published sweep against the same account, so two
+# pipelines can select the same AMI before either has deleted it. This handles
+# expected potential errors that could occur on a collision.
+delete_tolerating_missing() {
```
Alphasite: nit: should this be named something like: `run_tolerate_missing_errors` or something that doesn't specifically mention `delete` since its also used for deregister.

Alphasite (dismissed):
i *think* its functionally fine, but have a few minor nits. My main concern is around snapshot_id, but i figure you did your due diligence

ci/tasks/light-aws/cleanup-ami.sh
ystros: Its usage was removed https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/c2e1e3c02026b7fab1a6e94935041e8007acb931 about a month ago. Searching all pipeline definitions in Concourse showed no usages of `snapshot_id`.

beyhan:
During the FI WG meeting we wasn't sure whether this is ready for merge. @Alphasite you have some small improvement suggestions but also approved it. Is ready?

ystros:
Thanks! Just addressed the feedback, could you take another look @Alphasite ?

Alphasite (approved):

Alphasite:
Ah yes sorry. Approved LGTM. My concerns were relatively minor

PR #715, opened 2026-08-25
Make base_ubuntu_warden_rosetta host-architecture agnostic so it can run in CI

mkocher:
## Human Summary

Make building the rosetta stemcell possible in CI by not relying on any of the arm binaries. Verified with a build and test deployment with the new stemcell.

## AI Summary

The stage executed arm64 binaries while doing its work — the tar self-test, and `dpkg -x` in the chroot, which shells out to the just-swapped GNU tar — so it only ran on an Apple Silicon host. Download all the arm64 debs first, extract them from outside the chroot with the builder image's tar, then swap and statically verify (e_machine == 183) every replacement, including tar and the systemd daemons. The tar round-trip now runs only behind an exec probe and skips loudly otherwise, in the stage and in rosetta_spec.rb, so the stage can build on an x86-64 CI worker.

Also cover stage_collection's rosetta insertion, which had no unit test.


mkocher:
closing to reopen fresh against the right branch

PR #717, opened 2026-08-28
CI: replace usn git resource with partial-clone task

selzoc:
## Summary
- Replace the `(os)-usn-gh-json` git resource with a task
  (`fetch-usn-gh-json`) doing a partial clone (`--filter=blob:none
  --depth 1`) + cone sparse-checkout scoped to `usn/`, cached across
  builds via `caches:` so warm runs do an incremental `fetch` +
  `reset --hard` instead of a full clone.
- Root cause: the stock `git` resource does a plain `git clone`
  regardless of `sparse_paths`/`depth` config - sparse-checkout only
  trims the working-tree checkout, not what's fetched into
  `.git/objects`. `vex/` (23.8GB) and `osv/` (9.8GB) dominate the repo
  at HEAD; `usn/`, the only directory any of the 3 consuming tasks
  reads, is 570MB.
- Benchmarked cold on a real Concourse worker via `fly execute`
  (not a pipeline change):

  | | time | disk |
  |---|---|---|
  | old (plain clone) | 8909s (~2h28m) | 58.5GB |
  | new, cold | 43s | 664MB |
  | new, warm (cache hit) | 2s | - |

  ~207x faster cold, ~88x smaller, effectively free once cached.
- Dropped the resource's `passed: [process-high-critical-cves]` pin on
  `check-usn-packages-are-available` - that job's actual purpose is
  checking USN fix-availability against the *live* apt repo state, so
  always-fetch-latest is correct behavior, not a regression.
- Added `serial: true` to `check-usn-packages-are-available` to prevent
  concurrent builds racing on the same worker-scoped clone cache.

### DIFF
--- a/ci/pipeline-template.yml
+++ b/ci/pipeline-template.yml
@@ -280,7 +280,9 @@ jobs:
       - get: (@= data.values.stemcell_details.os_short_name @)-usn
         version: every
         trigger: true
-      - get: (@= data.values.stemcell_details.os_short_name @)-usn-gh-json
+      - task: fetch-usn-gh-json
+        file: bosh-stemcells-ci/ci/tasks/fetch-usn-gh-json.yml
+        image: os-image-stemcell-builder-registry-image
       - get: usn-log
       - get: high-critical-unfound-usns
     - task: check-if-usn-is-in-github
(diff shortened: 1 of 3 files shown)


KauzClay (approved):
nice

selzoc:
First test here: https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/ubuntu-jammy/jobs/process-high-critical-cves/builds/562

PR #718, opened 2026-08-28
CI: Move fetch-usn-gh-json task outside of in_parallel

selzoc:
Followup to #717 
It depends on a get of bosh-stemcells-ci in the in_parallel block, so it can't be parallelized with it.

Validated in the pipeline: https://bosh.ci.cloudfoundry.org/teams/stemcell/pipelines/ubuntu-jammy/jobs/process-high-critical-cves/builds/563


KauzClay (approved):
dang, the in parallel bit woulda been nice

PR #744, opened 2026-09-17
Run the bosh-agent rsyslog rule before the syslog forwarder

neddp:
## Summary

On Noble stemcells, bosh-agent's DEBUG output is forwarded to the central log drain for every VM, even though the stemcell ships an rsyslog rule intended to keep it on-box. This PR fixes the rule ordering so the `& stop` actually takes effect, keeping bosh-agent logs local.

## Background

Noble runs bosh-agent as a systemd unit (`StandardOutput=journal`, `SyslogIdentifier=bosh-agent`), so its output flows into rsyslog. The stemcell ships `rsyslog_90-bosh-agent.conf` to divert that to a local file and stop:

```
if $programname == 'bosh-agent' then
/var/log/bosh-agent.log
& stop
```

`/etc/rsyslog.d/` is processed in lexical order, and the syslog-release `syslog_forwarder` job drops in a catch-all forwarding rule at deploy time:

```
# 35-syslog-release-forwarding-rules.conf
*.* @@<drain>:5514;SyslogForwarderTemplate
```

Because `35-...` sorts before `90-bosh-agent.conf`, every bosh-agent message is forwarded to the drain **before** the `& stop` at rule 90 ever runs. The stop only prevents local re-logging at `50-default.conf`; it does nothing to prevent the forward.

## Impact

bosh-agent DEBUG-dumps the full `/proc/mounts` and `spec.json` on every ~30s heartbeat. On a host running many containers, `/proc/mounts` can be hundreds of overlay-mount lines, so this floods the central log store on Noble. (On Jammy, bosh-agent logs via svlogd to local files and never enters rsyslog, so it is unaffected; see Scope.)

## Change

- Rename `rsyslog_90-bosh-agent.conf` → `rsyslog_34-bosh-agent.conf` and install it as `/etc/rsyslog.d/34-bosh-agent.conf`, so the rule runs **before** the syslog-release forwarding rule (`35-...`) and the `& stop` keeps bosh-agent logs local.
- Add a comment on the rule documenting the ordering dependency.
- Update the stemcell specs (`google_spec.rb`, `stemcell_shared_examples.rb`) that assert the `/etc/rsyslog.d` file list.

## Scope: Noble only

This is intentionally Noble-only. On `ubuntu-jammy`, bosh-agent is started by runit and logs via `svlogd` to `/var/vcap/bosh/log`; there is no bosh-agent rsyslog rule, so its output never reaches rsyslog or the forwarder. The ordering bug cannot occur there, and a `& stop` rule would be a no-op. If Jammy is later migrated to the systemd/journal agent-logging model, it would need the same correctly-numbered rule.

## Prior work (the local-only logging intent)

Keeping bosh-agent logs on-box was the intent of earlier changes; this PR only fixes their ordering against the forwarder:

- c27910fd - migrate monit and bosh-agent to systemd (bosh-agent output now goes to the journal).
- ed2ff2a9 - stream/rotate the bosh-agent journal output to a file (`/var/log/bosh-agent.log`) instead of letting syslog write to `/var/vcap/bosh/log`; this is where `rsyslog_90-bosh-agent.conf` with `& stop` was introduced.
- #694 "fix: enable systemd forwarding to syslog" - makes systemd forward to syslog so bosh-agent logs flow through syslog to `/var/log/bosh-agent.log`.

The `& stop` in that rule was meant to keep the logs local, but its rule number (90) sorts after the syslog-release forwarding rule (35), so the forward wins. This PR corrects the ordering.

## Related

This is the transport-level half of the fix. The source-level changes that stop bosh-agent from dumping the content in the first place (and redact signed blobstore URLs) are separate bosh-agent / bosh-utils PRs; either fix alone reduces the volume, and together they address root cause plus this ordering defect.

## Testing

- Stemcell specs updated to expect `34-bosh-agent.conf`.
- After deploy, `/etc/rsyslog.d/34-bosh-agent.conf` precedes `35-syslog-release-forwarding-rules.conf`; bosh-agent entries land in `/var/log/bosh-agent.log` and are no longer forwarded to the drain.


ramonskie (approved):
i agree

